PR #66, opened 2015-06-03
Swapped out MyMapView.Extent for MyMapView.GetCurrentViewpoint(...)

macca400:


anttikajanus:
@macca400 

I think that it would be better to split `MyMapView.GetCurrentViewpoint(ViewpointType.BoundingGeometry).TargetGeometry.Extent` into separate lines to make it more clear what's going on. It would make also the lines a bit more readable (shorter). In some cases like 

````CSharp
// Get current viewpoints extent from the MapView
var currentViewpoint = MyMapView.GetCurrentViewpoint(ViewpointType.BoundingGeometry);
var extent = viewpoint.TargetGeometry.Extent;
````

katydalton:
@macca400 - Just one minor issue in the Desktop version of GraphicsSourceSample - on line 85 there's an extra "v" character that needs removed.

Otherwise good to merge.

PR #67, opened 2015-06-15
Kdalton feature layer edit attribute

katydalton:
@anttikajanus - Can you take a look? This is the fix we discussed with @pMaske via email 6/3/2015.


anttikajanus:
@katydalton looks good to me.

PR #75, opened 2015-06-24
Fixed local data paths to target AnyCPU

anttikajanus:
@katydalton Please review and merge if ok.

- Changed local path directories to target AnyCPU paths `..\..\..\..\` to `..\..\..\`
- Also fixed one of the urls in create featurelayers sample (code that is not really used). Do you think that the code is ok to be there or should I remove it totally. FeatureLayers are created in XAML and the code in the code behind is commented out but used to show how to do it there.


katydalton:
Looks good. I see what you mean about the Create FeatureLayer from Service sample. I'd say we leave it in, but maybe I'll update the comment to clarify that we're creating the feature layer in Xaml but you could also do it in code-behind?

anttikajanus:
Thanks. One thing to look when we go to through the samples is how much we have this kind of stuff.

PR #81, opened 2015-07-06
Updating to use a SceneService from scene.arcgis.com. Created new way…

katydalton:
@anttikajanus - Can you take a look? Just updated to use a service off scene.arcgis.com instead of the other server. Generated new waypoints for Vancouver. Note no textures on these services, but working on getting one.


anttikajanus:
@katydalton Looks good.

PR #82, opened 2015-07-07
Updated to use NavigationChanged instead of SpatialReferenceChanged.

katydalton:
@anttikajanus - Do we need to unload the NavigationChanged event handler before switching between samples on this one?


anttikajanus:
Looks good. Please merge.

PR #84, opened 2015-07-10
Kdalton FL from gdb

katydalton:
@mbranscomb - Was this roughly what you were thinking? Based on the FeatureLayer from GDB 2d sample. Can discuss more in other issue if needed.


katydalton:
Closing and will open new PR against Master now that we've merged.

PR #85, opened 2015-07-14
Adding new KML Samples

AsserSwelam:
Some of new KML Samples demonstrating the way to use it.


AsserSwelam:
@katydalton Please review

PR #89, opened 2015-08-09
Update kml service; rename to 3d instead of 3D.

katydalton:
@anttikajanus -Can you take a quick look. Just swapped out the service and did a minor rename.


anttikajanus:
@katydalton looks good.

PR #93, opened 2015-09-15
Move token security samples

katydalton:
@mbranscomb - Can you take a look. This branch includes changes to all three platforms to move Token Secured Services to separate projects. I ended up separating them into Challenge method and Known User method so we didn't have any commented-out code. These changes also include the addition of the SecuritySamples landing/placeholder page. And also moved OAuthAuthorizeHandler into the Online Services folder as a temporary measure until we move those samples out to their own projects under Security. @ThadT did a first-pass review to confirm new projects are working as expected.


mbranscomb:
@katydalton If you modify the user/password conditional operator in both app samples then I think it's good to go.

PR #94, opened 2015-09-15
Add KML Popup sample for Desktop

AsserSwelam:
Add KML Popup sample for Desktop


katydalton:
Looks good - okay to merge.

PR #102, opened 2015-10-21
Initial commit of OAuth samples for Desktop, Store, Phone. 

katydalton:
On disk only; not included with Solution file.

@mbranscomb - Could you take a quick look?


katydalton:
Updated to use NuGet. Modified/fixed text in Readme files. Updated Output location to match larger solution.

PR #190, opened 2016-07-28
Update to Author Map sample: save thumbnail with map

ThadT:
The WPF and UWP samples have been updated to include code that saves the current map view display as the portal item's thumbnail image. This code runs when a portal item is initially created (map is first saved) and each time it's updated.


ThadT:
@dkemlage Assigned to you for review ... don't worry about it if you don't have time. Two new functions were added to ach version of the author map sample (UpdatePortalItemAsync and WriteThumbnailImageAsync).

dkemlage:
@ThadT - I tested the C# and VB.NET versions of the WPF and UWP samples. They ran great - I can see the Thumbnail in the ArcGIS Online Account. The code looks good - no obvious issues. Good luck with the other SDK versions (iOS, Android, Forms) Check it in. -don

PR #222, opened 2017-01-05
WPF version (C# and VB) of the webmap tutorial

ThadT:
Hey @dkemlage 
If you have time, can you give this a review? It's basically the solution to the 'author, edit, and save maps to your portal' tutorial. I had to make it a little different to work in the sample viewer.


dkemlage:
@ThadT Here are my comments for the code review:

Both the CSharp and VB.NET versions of the sample worked well, without error. Congratulations!

One of my primary goals was to ensure that the tutorial guide document 'Author, edit, and save maps to your portal' (https://developers.arcgis.com/net/latest/wpf/guide/author-edit-and-save-maps-to-your-portal.htm) matched what was being shown in the sample code in the SampleViewer. So with that in mind:

1. There is no 'New Map' button in the UI screenshots nor is it wired up in the constructor of the tutorial document.

2. There are several using statements that can be eliminated (unused) in the C# code (they are fine in the VB.NET code):
using Esri.ArcGISRuntime.Geometry;
using System.IO;
using System.Windows.Media;
using System.Windows.Media.Imaging;

3. In the tutorial document the user is told to create a MapViewModel.cs file and add code to it. But in the actual sample in the SampleViewer, the MapViewModel is a sub-class of AuthorEditSaveMap class. This inconsistency should probably be resolved or at lest noted in the documentation. 

4. The actual code in the SampleViewer makes use of a _map variable in the MapViewModel class but it is not used in the tutorial.

5. Add some blank lines for readability in the sample code of the SampleViewer for the 'UpdateAuthenticationManager' function. The comments and actual code are bunched together.

6. There are some 'TODO:' comments in the VB.NET code that are not present in the CSharp code. This should be reconciled.

7. The order of the MapViewModel and OAuthAuthorize classes are different in the CSharp and VB.NET sample code. They should probably both flow the same when you read them.

8. The VB.NET code has a few sections of triple slash (///) comments with 'summary'  and .param' tags that do not exist in the CSharp code. This should be reconciled.

9. Some global (aka. member) variables do not begin with underscores ( _ ) in the AuthorEditSaveMap and they probably should for consistency:
Private Const ArcGISOnlineUrl As String = "https://www.arcgis.com/sharing/rest"
Private Const AppClientId As String = "2Gh53JRzkPtOENQq"
Private Const OAuthRedirectUrl As String = "https://developers.arcgis.com"

PR #234, opened 2017-02-17
Android version of the Manage Bookmarks sample

ThadT:
The Android version was the only one missing for this sample.

@macca400 
Please give this a look when you have a chance (should be an easy one?). Thanks!


ThadT:
Hey Don, assigning this to you since @macca400 is traveling today. Can you take a look today so we can get it in the doc update?

dkemlage:
@ThadT The code looks good. Check it into master!

PR #241, opened 2017-03-21
Don1760/three xamarin forms samples

dkemlage:
@macca400 Please code review for inclusion into the ArcGIS-runtime-samples-dotnet repo. The following Geoprocessing samples existed in WPF and UWP and I have ported them into the Xamarin Forms platform:
~ AnalyzeHotspots
~ AnalyzeViewshed
~ ListGeodatabaseVersions

I tried to keep the code the same as much as possible across the SDK's but there are some small platform UI differences that needed some liberties.


src/Forms/Shared/Samples/Geoprocessing/AnalyzeHotspots/AnalyzeHotspots.cs
```diff
@@ -0,0 +1,128 @@
...
+using Xamarin.Forms;
+
+namespace ArcGISRuntimeXamarin.Samples.AnalyzeHotspots
+{
+    public partial class AnalyzeHotspots : ContentPage
+    {
+
+        // Url for the geoprocessing service
+        private const string _hotspotUrl =
+            "http://sampleserver6.arcgisonline.com/arcgis/rest/services/911CallsHotspot/GPServer/911%20Calls%20Hotspot";
```
macca400: Any url used in an iOS sample is required to be https. Or if there is no https equivalent available then the domain of the url must be added to the info.plist

PR #245, opened 2017-04-07
Sketch on map sample for all platforms

ThadT:
Workflow sample with the following functionality:
 - Sketch all geometry types (sketch modes) and add them as graphics to the map view.
 - Select graphics to edit (move vertices, etc.).
 - Undo, redo, cancel, complete.
 - Clear graphics.


src/iOS/Xamarin.iOS/Samples/GraphicsOverlay/SketchOnMap/SketchOnMap.cs
```diff
@@ -0,0 +1,350 @@
...
+
+        // Segmented control to show sketch controls
+        private UISegmentedControl _segmentButton;
+        
+        public SketchOnMap()
+        {
+            Title = "Sketch on map";
+        }
+
+        public override void ViewDidLoad()
```
macca400: I added functionality today to allow iOS samples to rotate to adapt to device orientation. Having the CreateLayout() method call in the ViewDidLoad methods means that  when the device's orientation is changed the map will not change in size to fit the screen. 

Adding the CreateLayout() and Initialise() calls to a ViewDidLayoutSubviews() method instead will resolve this issue.

See https://devtopia.esri.com/runtime/dotnet-sdk/issues/1110 for details

macca400:
@ThadT  looks good man. 

One small change for ios sample and it's good to go!

PR #251, opened 2017-05-18
Reverted to ViewDidLoad, used ViewDidLayoutSubviews for Frame update

jNery:


jNery:
@macca400 : Merging the same fix. I left out one sample.

PR #260, opened 2017-06-15
J nery/offline map task xamarin forms

jNery:
Please review - samples based on WPF samples created by @anttikajanus


macca400:
Discussed with Antti removing PR to allow v.next to be deleted. Will create new PR against Master

PR #263, opened 2017-06-28
Offline map task WPF samples

macca400:


macca400:
Discussed with Antti removing PR to allow v.next to be deleted

PR #266, opened 2017-07-13
Sample geocode find address

nCastle1:
This is an implementation of the Find Address sample from common-samples.


nCastle1:
PR was closed and #275 was opened as that was the most straightforward way to undo the damage of my mangled commit-squash attempt.

PR #268, opened 2017-07-14
Cosmetic fix map layer forms

nCastle1:
This resolves an issue where the ‘display layer view state’ sample did not render appropriately on Xamarin.Forms.


nCastle1:
Note that the solution file was modified because the WPF C# sample was not appearing in the right folder. At the time I fixed that, I thought it was the result of an errant drag-and-drop on my branch. Further review revealed that the WPF C# sample is in the wrong folder on master as well.

nCastle1:
See that the edit to the solution file was undone. That will be put into a separate branch/issue/pr

nCastle1:
@dkemlage Can you please take a look at this when it's convenient?

src/Forms/Shared/Samples/MapView/DisplayLayerViewState/DisplayLayerViewState.xaml
```diff
@@ -7,9 +7,9 @@
   <Grid>
     <Grid.RowDefinitions>
       <RowDefinition Height="*" />
-      <RowDefinition Height="auto" />
+      <RowDefinition Height="100" />
```
dkemlage: Lets bump up the height of the listbox to say 150 so that you can see all three line items without having to scroll.

dkemlage:
@nCastle1 Small UI size change, then feel free to Merge.

PR #281, opened 2017-07-21
Update Info.plist to allow insecure content

nCastle1:
Many webmaps feature http content and we need that to show up on iOS.
The exceptions were copied from the iOS (native) SDK for the ArcGIS
Runtime.


nCastle1:
@dkemlage Can you take a look at this when you get a free moment?

dkemlage:
The file changes look good. I was able to confirm that the sample 'Open Existing Map' is now able to display the 'USA Tapestry Segmentation' layer:

image

Please merge code.

PR #287, opened 2017-07-26
Open Mobile Map (mmpk) sample

nCastle1:
I'm opening this PR to track work related to the sample for opening a mobile map from an mmpk.


nCastle1:
At this point most platforms are done, but Android is still WIP.

nCastle1:
@dkemlage Can you take a look at this when you get a moment? Pending any review suggestions, I believe this is ready to go.

src/Forms/Shared/Samples/Map/OpenMobileMap/OpenMobileMap.xaml.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            Title = "Open mobile map (map package)";
+
+            // Create the UI, setup the control references and execute initialization
+            Initialize();
+        }
+
+        private async void Initialize()
+        {
+            // Because each platform handles resource embedding differently,
```
dkemlage: Wow, this section of code (next 20 lines or so) is brilliant! Well done using the conditional if/def statements. My only comment here that you add more comments of what is occurring in each platform. Are there possibly any Xamarin or Microsoft documents links that you could provide on the syntax differences for each platform for getting the assembly and folder in order to construct the path to the .mmpk?

src/Forms/Shared/Samples/Map/OpenMobileMap/OpenMobileMap.xaml.cs
```diff
@@ -0,0 +1,85 @@
...
+#elif __ANDROID__
+			var resourcePrefix = "ArcGISRuntimeXamarin.";
+			var assembly = this.GetType().Assembly;
+			var folder = Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments);
+#endif
+            // The path on disk for the file
+            var path = Path.Combine(folder, mmpkName);
+
+            // Copy the file to disk if it isn't already there
+            if (!File.Exists(path))
```
dkemlage: Again I would like to see more comments of what is actually happening in the code regarding opening the .mmpk so that it can be read using the MobileMapPackage.OpenAsycn(path).

nCastle1:
I'm closing this PR per Mike's feedback. I'll open another PR once I've implemented the correct approach to offline data for this sample.

PR #290, opened 2017-08-01
Fixed duplicate file include in iOS sample project

fragmer:
ArcGISMapImageLayerUrl.cs was listed twice in csproj file (lines 354 and 359), causing a build error.


nCastle1:
Looks good.

PR #296, opened 2017-08-08
Update sample viewer to skip unimplemented samples

nCastle1:
The sample viewer needs to be updated as there are now samples that are implemented only for C# (and not VB). This PR accomplishes that by adding a new property to the `SampleModel` called `SampleVbTypeName` which provides the assembly-qualified name of the VB sample (should it exist). This is then passed to `System.Type.GetType` on sample creation; if the result is null, the sample model is not created.

### DIFF
--- a/src/ArcGISRuntime.Samples.Shared/Models/SampleModel.cs
+++ b/src/ArcGISRuntime.Samples.Shared/Models/SampleModel.cs
@@ -14,7 +14,6 @@ using System.IO;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Json;
 using System.Text;
-using System.Reflection;
 
 namespace ArcGISRuntime.Samples.Models
 {
--- a/src/ArcGISRuntime.Samples.Shared/Models/SampleModel.cs
+++ b/src/ArcGISRuntime.Samples.Shared/Models/SampleModel.cs
@@ -7,12 +7,14 @@
 // "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific 
 // language governing permissions and limitations under the License.
 
+using ArcGISRuntime.Samples.Managers;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Json;
 using System.Text;
+using System.Reflection;
 
 namespace ArcGISRuntime.Samples.Models
 {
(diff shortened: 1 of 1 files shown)


nCastle1:
@macca400 Do you have some time to check this out? This approach seemed straightforward to me, but I'm very interested if there is a better way (I understand reflection is expensive) and you know the code better than I. Thanks!

src/ArcGISRuntime.Samples.Shared/Models/SampleModel.cs
```diff
@@ -204,9 +222,15 @@ internal static SampleModel Create(string metadataFilePath)
                 sampleModel.SampleFolder = metadataFile.Directory;
             }
 
+            // Stop if the sample doesn't have a VB implementation (VB sample viewer only)
+            if (ApplicationManager.Current.SelectedLanguage == Language.VBNet && System.Type.GetType(sampleModel.SampleVbTypeName, false) == null)
```
macca400: You have the sample's folder available, could you just check the \<samplename\>VB.xaml exists? if not return

src/ArcGISRuntime.Samples.Shared/Models/SampleModel.cs
```diff
@@ -141,6 +143,22 @@ public enum SampleType
         public DirectoryInfo SampleFolder { get; set; }
 
         /// <summary>
+        /// Gets the (expected) assembly-qualified name for the VB sample
+        /// </summary>
+        [IgnoreDataMember]
+        public String SampleVbTypeName
```
macca400: Without the summary comment of the property I wouldn't know what SampleVbTypeName was for

src/ArcGISRuntime.Samples.Shared/Models/SampleModel.cs
nCastle1: I've investigated this and it turns out the XAML files aren't included in the output directory, so that can't be used reliably to check that the sample was implemented. Because most of the properties in the sample model are taken from metadata.json (which is shared between the two types) I think reflection is the only way to go. :/

nCastle1:
@macca400 I've tried to address as many of your concerns as I could. Would you mind taking one last look to confirm that the changes are satisfactory before I merge? Thank you! :)

macca400:
Feel free to merge this @nCastle1 looks good.

One last though how about the ExpectedVbAssemblyQualifiedType property was AssemblyQualifiedType and it had a logic to decide to whether to return the VB or C# assembly? 

It would be nice to keep the property generic. Who knows maybe at some point we'll have a VB implementation and want to ignore a c# sample.

nCastle1:
I'm not sure it would be possible to do checking to see which assembly type is desired within a property. I agree it would be nice to have it be generic, but given what is known at run time I don't think its practical. I'm merging this but will make a mental note to investigate that.

PR #307, opened 2017-09-12
Adding the 'Analyze Viewshed' sample for Xamarin.iOS to match parity with other .NET SDK platform samples.

dkemlage:
@nCastle1 Can you please do a code review on this sample: 'Analyze Viewshed'. All of the other .NET SDK versions (WPF, UWP, Xamarin.Forms, and Xamarin.Android) of this sample are already done. Just doing some backfilling for Xamarin.iOS since I have a little extra time. Thanks, -don


src/iOS/Xamarin.iOS/Samples/Geoprocessing/AnalyzeViewshed/AnalyzeViewshed.cs
```diff
@@ -0,0 +1,212 @@
...
+                }
+                else
+                {
+                    // Report error
+                    UIAlertController alert = UIAlertController.Create("Sample Error", ex.ToString(), UIAlertControllerStyle.Alert);
+                    alert.AddAction(UIAlertAction.Create("OK", UIAlertActionStyle.Default, null));
+                    PresentViewController(alert, true, null);
+                }
+            }
+            finally
```
nCastle1: I think we should get rid of the finally block. Because we're not doing anything here, I think it might confuse users.

PR #309, opened 2017-09-14
Statistical query with group and sort of results

ThadT:
Complete for all platforms.


ThadT:
Thanks @nCastle1! 
This is a lot of code, please take your time looking this over. Don't be shy about making lots of suggestions/comments. I'm happy with the sample, but not sure it's the most efficient/consice way to do this.

src/Android/Xamarin.Android/Samples/Data/StatsQueryGroupAndSort/StatsQueryGroupAndSort.cs
```diff
@@ -0,0 +1,1000 @@
...
+            groupFieldsDialog.Show(trans, "group_flds");
+        }
+
+        private void ShowOrderByFields(object sender, EventArgs e)
+        {
+            // If there are no available order fields, don't show the (empty) list
+            if (_orderByFields.Count == 0)
+            {
+                // Warn the user to choose group fields first
+                var alertBuilder = new AlertDialog.Builder(this);
```
nCastle1: This is just a suggestion, but I like the pattern of having a separate 'ShowMessage' function. This way the code can be kept more consistent between samples and any repetition is avoided when multiple functions need to show messages.

src/Android/Xamarin.Android/Samples/Data/StatsQueryGroupAndSort/StatsQueryGroupAndSort.cs
```diff
@@ -0,0 +1,1000 @@
...
+            // Dialog UI to display
+            LinearLayout dialogView = null;
+
+            // Get the context for creating the dialog controls
+            Android.Content.Context ctx = this.Activity.ApplicationContext;
+
+            // Set a dialog title
+            this.Dialog.SetTitle("Statistics Definitions");
+
+            try
```
nCastle1: Is this UI code likely to fail? If not, it might be clearer to get rid of the try/catch.

src/iOS/Xamarin.iOS/Samples/Data/StatsQueryGroupAndSort/StatsQueryGroupAndSort.cs
```diff
@@ -0,0 +1,847 @@
...
+
+        // Model for defining choices in the statistics definition UIPickerView
+        private StatDefinitionModel _statsPickerModel;
+
+        // List of statistics definitions to use in the query
+        private List<StatisticDefinition> _statisticDefinitions = new List<StatisticDefinition>();
+
+        public StatsQueryGroupAndSort()
+        {
+            Title = "Stats query with grouped results";
```
nCastle1: This title is inconsistent with the titles in the other samples.

ThadT:
Thanks @nCastle1, all good suggestions. I'll address these comments and let you know when it's ready for another look.

ThadT:
OK @nCastle1 
I think I addressed everything across all platforms. Give it a try when you have a chance and let me know if it needs anything else. (I'll address the conflicts with vnext)

nCastle1:
@ThadT I think there might be something wrong with the repo... I see 677 files changed.

nCastle1:
@ThadT please disregard, it looks like that was due to the merge.

PR #310, opened 2017-09-14
WMS Layer (URL)

nCastle1:
This is the implementation of the WMS layer url sample.


nCastle1:
@ThadT I've requested your review on this. Thank you! :)

PR #311, opened 2017-09-16
Don1760/raster rendering rule new sample

dkemlage:
@nCastle1 Can you please code-review this new .NET sample called 'RasterRenderingRule'?
I will be out on vacation for 2 weeks, so if you would be so kind as to make whatever code-review changes you feel are necessary and merge it into master, I would appreciate it. I will co-assign the issue to you as well so you can get credit for the work you did.
Thanks,
-don


src/Android/Xamarin.Android/Samples/Layers/RasterRenderingRule/RasterRenderingRule.cs
```diff
@@ -0,0 +1,173 @@
...
+
+namespace ArcGISRuntimeXamarin.Samples.RasterRenderingRule
+{
+    [Activity]
+    public class RasterRenderingRule : Activity
+    {
+        // Create and hold reference to the used MapView
+        private MapView _myMapView = new MapView();
+
+        // Create a global scope empty read-only list for the various rendering rules of the image service raster
```
nCastle1: I don't like the wording of 'global scope' here. The object is scoped to the class and I don't think 'global' feels right in the context of .NET/C#.

src/Android/Xamarin.Android/Samples/Layers/RasterRenderingRule/RasterRenderingRule.cs
```diff
@@ -0,0 +1,173 @@
...
+            // Set the Uri for the image server
+            _myUri = new Uri("https://sampleserver6.arcgisonline.com/arcgis/rest/services/CharlotteLAS/ImageServer");
+
+            // Create a new image service raster from the Uri
+            ImageServiceRaster myImageServiceRaster = new ImageServiceRaster(_myUri);
+
+            // Load the image service raster
+            await myImageServiceRaster.LoadAsync();
+
+            // Get the ArcGIS image service info (aka. metadata) from the image service raster
```
nCastle1: nitpick: I don't know if "aka." is necessary here.

src/Android/Xamarin.Android/Samples/Layers/RasterRenderingRule/RasterRenderingRule.cs
```diff
@@ -0,0 +1,173 @@
...
+        private async void Initialize()
+        {
+            // Assign a new map to the MapView
+            _myMapView.Map = new Map();
+
+            // Set the basemap to Streets
+            _myMapView.Map.Basemap = Basemap.CreateStreets();
+
+            // Set the Uri for the image server
+            _myUri = new Uri("https://sampleserver6.arcgisonline.com/arcgis/rest/services/CharlotteLAS/ImageServer");
```
nCastle1: The pattern I've seen in other samples is to move the Uri to a class-level variable. I don't know if that needs to change in this case but maybe there should be a style guide for such things?

src/Android/Xamarin.Android/Samples/Layers/RasterRenderingRule/RasterRenderingRule.cs
```diff
@@ -0,0 +1,173 @@
...
+            // Define a new view point from the full extent envelope
+            Viewpoint myViewPoint = new Viewpoint(myEnvelope);
+
+            // Zoom to the area of the full extent envelope of the image service raster
+            await _myMapView.SetViewpointAsync(myViewPoint);
+
+            // Get the rendering rule info (i.e. definitions of how the image should be drawn) info from the image service raster
+            _myReadOnlyListRenderRuleInfos = myArcGISImageServiceInfo.RenderingRuleInfos;
+
+            // Make sure that we have at least one render rule info to proceed
```
nCastle1: In this particular case, I don't think we need to check for 0 rules. In that case, the code in the `foreach` loop won't execute, which is consistent with the behavior with the `Count > 0` check.

src/Android/Xamarin.Android/Samples/Layers/RasterRenderingRule/RasterRenderingRule.cs
```diff
@@ -0,0 +1,173 @@
...
+            // Get the rendering rule button
+            var renderingRuleButton = sender as Button;
+
+            // Create menu to show the rendering rule options
+            var renderingRuleMenu = new PopupMenu(this, renderingRuleButton);
+            renderingRuleMenu.MenuItemClick += OnChangeRenderingRuleMenuItemClicked;
+
+            // Create menu options
+            foreach (var renderingRuleName in _names)
+                renderingRuleMenu.Menu.Add(renderingRuleName);
```
nCastle1: Although this is valid code, I really, really don't like omitting the curly brackets. As written, the indentation may imply to less experienced developers that the indentation is meaningful (like in Python). This may cause pain for users.

src/Android/Xamarin.Android/Samples/Layers/RasterRenderingRule/RasterRenderingRule.cs
```diff
@@ -0,0 +1,173 @@
...
+            OnChangeRenderingRuleClicked(_renderingRulesButton, null);
+        }
+
+        private void OnChangeRenderingRuleClicked(object sender, EventArgs e)
+        {
+            // Get the rendering rule button
+            var renderingRuleButton = sender as Button;
+
+            // Create menu to show the rendering rule options
+            var renderingRuleMenu = new PopupMenu(this, renderingRuleButton);
```
nCastle1: I don't like the use of `var` here and throughout the sample. Although var is very convenient, it doesn't read well in code listings outside of an IDE.

src/Android/Xamarin.Android/Samples/Layers/RasterRenderingRule/RasterRenderingRule.cs
nCastle1: After re-reading I see that this is a class-scope variable. I would prefer if this was initialized where it is declared rather than in a separate variable. If the user sees this referenced in another function, they will need to search for the code that assigns the value rather than just looking at the declaration.

src/UWP/ArcGISRuntime.UWP.Samples/Samples/Layers/RasterRenderingRule/RasterRenderingRule.xaml.cs
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }
+
+            // Set the combo box index to the first rendering rule info name
+            comboBox_RenderingRuleChooser.SelectedIndex = 0;
+        }
+
+        private void comboBox_RenderingRuleChooser_SelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            // This function executes when a rendering rule info name is changed in the combo box
```
nCastle1: I would prefer to move this to an xml comment on the function.

src/UWP/ArcGISRuntime.UWP.Viewer.sln
```diff
@@ -14,6 +14,7 @@ EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		..\ArcGISRuntime.Samples.Shared\ArcGISRuntime.Samples.Shared.projitems*{0ad85845-0ba1-4dc9-8c32-e6fd7632e32d}*SharedItemsImports = 13
+		..\ArcGISRuntime.Samples.Shared\ArcGISRuntime.Samples.Shared.projitems*{6820b615-c568-4a13-b026-a77cdd348bfa}*SharedItemsImports = 4
```
nCastle1: I don't know why this is here.

src/WPF/ArcGISRuntime.WPF.Viewer.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
```
nCastle1: I think we'll want to un-commit this change.

src/UWP/ArcGISRuntime.UWP.Viewer.sln
fragmer: Microsoft doesn't really know either: https://github.com/Microsoft/msbuild/issues/545

nCastle1: @fragmer I was planning on reverting this change. Do you see any problems with that?

fragmer: I'd just be weary of removing VS-generated changes to sln/csproj files, in case it breaks something in a subtle/obscure way. You'd be on the hook to test building every configuration for every platform in every VS version if you revert this ;)

nCastle1:
With changes this is ready to merge.

PR #318, opened 2017-09-27
KML layer (file)

nCastle1:


nCastle1:
@macca400 Can you take a look at this when you get a chance? It is *very* similar to the sample in the other PR.

nCastle1:
@macca400 I've assigned @dkemlage as he has some capacity right now and he worked on the design for these samples extensively. Thanks!

PR #319, opened 2017-09-27
Changes Xamarin Android package name

nCastle1:
With this change, it will now be possible to use the native and the
Forms versions of the sample viewer side-by-side on Android without errors. This brings the experience up to par with that of iOS.


nCastle1:
Hi @mbranscomb, GitHub suggested I ask you to review this. Feel free to unassign yourself if you don't have the time. 

This change was made because without it, it is challenging to work with both the Xamarin.Android and the Xamarin.Forms Android solutions simultaneously. Attempting to start one of the viewers while the other is installed would result in a `TypeLoadException` because the package names conflicted.

PR #322, opened 2017-10-03
Upgrades Xamarin.Forms to 2.4 stable

nCastle1:
Additionally, works around issues introduced in Xamarin.Forms 2.4, including bug 59868.


nCastle1:
@mbranscomb I've updated the nuget packages to match the new requirements for 100.2. Unfortunately, Xamarin.Forms 2.4 stable has some breaking changes that affected our samples. This PR includes fixes for both the build-breaking changes and the UI-breaking changes. 

Note that this work needs to be done now because the latest builds of 100.2 depend on (and cause nuget to automatically restore) XF2.4.0; without this PR, v.next will no longer build with the latest runtime.

fragmer:
...and a new version of Forms is out already! https://developer.xamarin.com/releases/xamarin-forms/xamarin-forms-2.4/2.4.0-sr1/

nCastle1:
@fragmer I saw that. I didn't upgrade the sampleviewer to that because it doesn't fix any of our bugs. I figured we keep it on the minimum required until there's a good reason not to. @mbranscomb what do you think?

fragmer:
@nCastle1 Yep you're right. We don't gain anything by going from "2.4.0.280+" to "2.4.0.282+".

PR #323, opened 2017-10-05
Change Time Extent

nCastle1:
Sample demonstrates how changing the time extent on the MapView causes time-aware layers to filter their data accordingly. 

Note that this is not ready for review yet; the chosen data sources are temporary. Once the data sources are changed to the final selection, a review will be requested.


nCastle1:
Closing this pull request; new one to be opened with updated sample.

PR #328, opened 2017-10-13
Add shapefile

ThadT:
Add a simple point shapefile from a local (downloaded) source.


ThadT:
Hi @nCastle1 
Can you give this a look when you have a chance (no hurry)? Thanks!

src/Android/Xamarin.Android/Samples/Data/AddShapefile/AddShapefile.cs
```diff
@@ -0,0 +1,88 @@
...
+            // Download (if necessary) and add a local shapefile dataset to the map
+            Initialize();
+        }
+
+        private async void Initialize()
+        {
+            // Create a new map to display in the map view with a streets basemap
+            _myMapView.Map = new Map(Basemap.CreateStreets());
+
+            // The shapefile will be downloaded from ArcGIS Online
```
nCastle1: There's a new process for offline data documented [here](https://github.com/Esri/arcgis-runtime-samples-dotnet/wiki/Offline-Data). I'm just realizing now that I never told anyone... 

Basically, the data manager code needs to go into a separate function with region tags. That makes it easier for the tooling to replace it with more appropriate code when injected into the API ref.

src/Forms/Android/ArcGISRuntime.Xamarin.Forms.Android.csproj
```diff
@@ -98,6 +98,9 @@
     <Reference Include="Validation, Version=2.4.0.0, Culture=neutral, PublicKeyToken=2fc06f0d701809a7, processorArchitecture=MSIL">
       <HintPath>..\..\..\packages\Validation.2.4.15\lib\netstandard1.3\Validation.dll</HintPath>
     </Reference>
+    <Reference Include="Windows.Foundation.UniversalApiContract">
```
nCastle1: It looks like this shouldn't have been included... I don't know if its necessary or will have side effects.

src/Forms/Shared/Samples/Data/AddShapefile/AddShapefile.xaml.cs
```diff
@@ -0,0 +1,68 @@
...
+            // Open a shapefile stored locally and add it to the map as a feature layer
+            Initialize();
+        }
+
+        private async void Initialize()
+        {
+            // Create a new map to display in the map view with a streets basemap
+            MyMapView.Map = new Map(Basemap.CreateStreets());
+
+            // The shapefile will be downloaded from ArcGIS Online
```
nCastle1: Same thing here re:offline data. My bad for not telling anyone

nCastle1:
I had some extra time so I made the changes per code review. Should be good to merge now.

PR #329, opened 2017-10-16
Add GeoPackage feature table

ThadT:
Hey @nCastle1 .. no hurry on this. I merged with v.next and ended up with over 600 changes in the PR. I think most of that was a few missing samples that have gone in over the last couple of days.


src/iOS/Xamarin.iOS/Samples/NetworkAnalysis/FindRoute/FindRoute.cs
```diff
@@ -199,7 +199,6 @@ private void ResetButtonClick(object sender, EventArgs e)
 
         private void ShowDirections(object sender, EventArgs e)
         {
-          //  ShowDirectionsController routeDirections = new ShowDirectionsController(_directionsList);
```
nCastle1: I don't know if you intended to include this.

src/Android/Xamarin.Android/Samples/Data/AddGeoPackageFeatureTable/AddGeoPackageFeatureTable.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        private async Task<string> GetGeoPackagePath()
+        {
+            #region offline data
+
+            // The GeoPackage will be downloaded from ArcGIS Online.
+            // The data manager (a component of the sample viewer), *NOT* the runtime handles the offline data process
+
+            // The desired GPKG is expected to be called "AuroraCO.shp"
```
nCastle1: It looks like this comment didn't get changed. If this is the expected behavior then there is no problem, otherwise this will need to change for each implementation.

PR #334, opened 2017-10-23
Added iOS version of the AnalyzeHotpots to the .NET public sample viewer. All the other .NET SDK versions of this sample were done previously.

dkemlage:
@nCastle1 
Would you please code review this new iOS code sample (AnalyzeHotspots). Thx, -don


dkemlage:
@nCastle1 Thank you for the code review.
I will merge as-is for this sample to save time. WPF and UWP use a date picker. However, Xamarin.Forms.UWP, Xamarin.Forms.iOS, Xamarin.Forms.Android, and Xamarin.Android use a simple text box for dates. So I will go with the path of least resistance. :)

PR #339, opened 2017-10-26
Identify WMS Features

nCastle1:


nCastle1:
@dkemlage Can you please take a look at this? It should be ready for review on all platforms. 

Note that there is an issue with callouts on Android that I'm investigating. I believe it is a bug and I'm working on a bug report. The sample still works, you'll just need to pan the map before the callout appears.

PR #345, opened 2017-11-06
Removes KML and updates readmes

nCastle1:


nCastle1:
@dkemlage Can you please take a look when you get the chance?

PR #348, opened 2017-11-07
Adding preliminary WPF sample for IdentifyKMLFeatures.

dkemlage:
@nCastle1 Would you please do a sanity code review for the WPF only version of the IdentifyKMLFeatures sample. It does not work yet since the API code is not wired up, but this sample follows the coding pattern that other layer types follow.


dkemlage:
Added Nathans comments in the .cs file so that ShowCalloutAt -vs- GraphicOverlays can be looked at when KML Identify is implemented in the future.

PR #354, opened 2017-11-16
Feature layer renderering mode map

macca400:
Sample code for FeatureLayer rendering mode (map)


macca400:
@dkemlage do you have time to have a look at this for me? Thought it made sense as you'll be able to do something very similar for the SceneView equivalent

macca400:
@dkemlage no problem, thanks for the review

PR #358, opened 2017-11-17
Begin enforcing C# language version

nCastle1:
Runtime supports customers using C# language version 5, so samples must not use newer language versions. Now, the project files will enforce that requirement. This change also removes the existing usages of C# 6 language features, including await in catch clauses, null propagation operator, and auto-property initializers.


nCastle1:
@macca400 Do you mind taking a look at this? There's no issue and I don't think we made plans to do the work, but I had some awkwardly-placed free time and I thought I'd take it on. Thanks! :)

dotMorten:
Just to be clear, since this PR caused some confusion among a few customers: The Runtime has no issue with anyone using the latest language features. All this PR is doing is to ensure our own samples doesn't accidentally use the latest language features, so customers still on older VS versions can copy-paste the sample code as-is and still compile on any supported VS version.

src/Android/Xamarin.Android/Samples/Layers/WmsServiceCatalog/WmsServiceCatalog.cs
```diff
@@ -187,6 +187,7 @@ public class LayerDisplayVM
 
         public LayerDisplayVM(WmsLayerInfo info)
         {
+            IsEnabled = false;
```
dotMorten: This line is unnecessary.

src/Forms/Shared/Samples/Data/FeatureLayerQuery/FeatureLayerQuery.xaml.cs
```diff
@@ -123,9 +123,9 @@ private void Initialize()
                     await DisplayAlert("State Not Found!", "Add a valid state name.", "OK");
                 }
             }
-            catch (Exception ex)
+            catch (Exception)
             {
-                await DisplayAlert("Sample error", "An error occurred", "OK");
+                DisplayAlert("Sample error", "An error occurred", "OK");
```
dotMorten: Isn't this generating a warning? (same for multiple features below)

src/Forms/Shared/Samples/Layers/WmsServiceCatalog/WmsServiceCatalog.xaml.cs
```diff
@@ -154,6 +154,7 @@ public class LayerDisplayVM
 
         public LayerDisplayVM(WmsLayerInfo info)
         {
+            IsEnabled = false;
```
dotMorten: Unnecessary change

nCastle1:
Changes made per PR, merging

PR #362, opened 2017-11-29
Updating the screen shots for the samples for consistency.

dkemlage:
Having screen shots less than 800 pixels wide is a best practice for display in web pages. Also using the same emulator gives a consistent look.


dkemlage:
@nCastle1 can you please do a code review (rather a screen shot review). Thx -don

PR #365, opened 2017-12-05
Certification fixes: round 2

nCastle1:
Resolves the last few lingering certification issues that were identified.


nCastle1:
@dkemlage, I have good news: all but one certification issue have been fixed. This small PR includes the following changes:

* fixes an issue where the UI was difficult to use for a sample on Xamarin.Forms UWP
* improves the UI on Find Place so that any user-placed callout disappears when a new search starts
* improves WMS layer URL to add the map to the map view after its initial viewpoint has been set (it doesn't make sense to set an initial viewpoint after the map is displayed). 
* fixes an issue that prevented Xamarin.Forms projects from loading

Can you please do a quick code review for me?

nCastle1:
Thanks for the review!

PR #372, opened 2017-12-11
Fix for canceling authentication on mobile

nCastle1:
There were issues happening when users cancelled authentication on the Xamarin platforms. These have been addressed by this PR.


nCastle1:
@dkemlage Can you please take a look at this? These issues were revealed by the resolution of a bug in authentication manager. You should be able to cancel any authentication in progress and then be able to try the operation again and successfully authenticate. 

Affected samples are Author and Save a Map, Author Edit and Save a Map, and Search a Portal for Maps.

nCastle1:
@ThadT It looks like Don will be at jury duty for the next day (or possibly longer). GitHub says you're the right person to ask for a code review. Do you mind giving this a review?

As mentioned above, the issues were related to canceling authentication. Samples where the user canceled the authentication, either with the cancel button or the close button, would break authentication until the app or sample was restarted. This should address these problems.

nCastle1:
Thanks for the review. Much appreciated :)

PR #373, opened 2017-12-11
Change android http client implementation

nCastle1:
This was prompted by a change to the version being used by the .NET api


nCastle1:
@dkemlage Can you please review this? Thanks! :)

mbranscomb:
@nCastle1 We should investigate updating the security samples too. This change should work in most cases, but I believe IWA is not supported by the Android handler.

nCastle1:
Update: working with Don, we found that things are equally broken on v.next, which is concerning.

nCastle1:
As discussed above, this appears to be an unrelated issue, possibly related to the Visual Studio emulator. I tested on Android 6, 4.2.2, and an HTC U11, none of which show the same issue. We'll want to make an issue to investigate this specific problem. 

The security samples have now been updated as well, with the exception of the IWA projects.

PR #375, opened 2017-12-11
Make WPF viewer behave when runtime components are missing

nCastle1:
WIll show a message and shut down if required components are not installed.


nCastle1:
@dkemlage can you please take a look at this? This resolves an issue where the application crashes without warning if run on devices that don't have the C++ redistributable installed.

PR #376, opened 2017-12-14
Clear Analysis results before new run

nCastle1:
Results of older analysis runs were not being cleared from the map
before new results were being shown.

This affects the 'Analyze Hotspots' sample only.


nCastle1:
@dkemlage  Can you please give this a review when you get a chance?

nCastle1:
@dkemlage Thanks for the review :)

PR #377, opened 2017-12-14
Update Xamarin.Forms UWP assets

nCastle1:
This change brings the Xamarin.Forms UWP project in line with all other
samples viewers in terms of visuals.

This change was made by copy-pasting the assets from the native UWP
project. Tested on desktop and everything looked right.


nCastle1:
@dkemlage Can you please take a look? This should be a purely aesthetic change. Thanks! :)

nCastle1:
Thanks for the review! :)

PR #378, opened 2017-12-14
Improves GeoViewSync

nCastle1:
This change disables the 'flick' gesture for both map and sceneview.
This gesture was problematic, because it was possible for the user to
start navigating on the scene, only to be fought when attempting to
navigate the map - the sceneview was still actively navigating.

With this change, the sample is now demonstrating similar behavior to
what is seen in ArcGIS pro when a map and a scene are linked.


nCastle1:
@dkemlage Can you please take a look? The behavior you should see is that if you try to 'fling' the scene or map in a particular direction that movement stops when you release the mouse button.

nCastle1:
Thanks for the review! :)

PR #379, opened 2017-12-14
Adds elevation source to scene layer sample

nCastle1:
Previously there was no elevation source, so buildings appeared to be floating. Now, all buildings should appear flush with the ground (generally). This also makes the sample look nicer as the ground looks much more realistic (and accurate).

Note that automated code cleanup was run, so there are some extra
formatting changes.


nCastle1:
@dkemlage Can you please take a look if you have a chance? I know I've asked a lot so if you don't have capacity I can ask someone else. Thanks! :)

nCastle1:
Thanks for the review! :)

PR #380, opened 2017-12-14
Normalize geometry when querying

nCastle1:
When wraparound is enabled on a map, the coordinates returned on tap are different when the map has been wrapped around. Because of this, panning the map and attempting to do a query will result in service exceptions.

Using the geometry engine to normalize the coordinates will prevent errors.

This only applies to situations where wraparound is enabled. Because of that, I chose to explicitly check if it is enabled, even though that is the default. That was a didactic decision rather than a practical one, but I think it is important for clarity. I did choose to put the condition check and normalization on one line;  this may be controversial, but I didn't want to divert emphasis from the intention of the sample.


nCastle1:
@dkemlage Can you please take a look at this? The errors caused by this behavior are unlikely to be encountered in the general operation of the samples. However, it was important to get this in because the Runtime behavior isn't intuitive and we don't document this anywhere else.

nCastle1:
Thanks for the review! :)

PR #382, opened 2017-12-14
Allows 'return' to dismiss keyboard on iOS

nCastle1:
Previously, it was not possible to dismiss the keyboard once it was opened in a sample. This makes it very difficult to use samples on device or when there isn’t a hardware keyboard attached.

With this change, pressing ‘return’ will hide the keyboard. This will enable our samples to be used on device in addition to just on the
simulator.


nCastle1:

nCastle1:
Thanks for the review! :)

PR #383, opened 2017-12-14
Fixes button in Viewshed Camera Sample

nCastle1:
The view shed camera button was using the ‘TouchUpOutside’ event which
wasn’t the correct event. Instead, ‘TouchUpInside’ should be used.

I also took the liberty of fixing the sample to handle rotation as well, so the code changes are a little larger than just the
Outside->Inside change.


nCastle1:

nCastle1:
Thanks for the review! :)

PR #384, opened 2017-12-14
Removes problematic ToolbarItems on Forms

nCastle1:
ContentPage.ToolbarItems has proven to be problematic on Xamarin.Forms. Toolbar items are not rendered consistently between platforms, making it difficult to tell that there is a control present on UWP.

A previous effort to remove instances of ContentPage.ToolbarItems was not entirely successful. This change fixes that.


nCastle1:

nCastle1:
Thanks for the review! :)

PR #385, opened 2017-12-14
Sets viewpoint consistently for raster layer file

nCastle1:
The Raster Layer File sample was having an issue where the viewpoint
wasn't being set consistently on load. There were also reports of a
crash that may have been due to a race condition. That issue should be
fixed as well.


nCastle1:

nCastle1:
Thanks for the review! :)

PR #386, opened 2017-12-15
Improves callout experience on Android

nCastle1:
The WMS Identify sample shows a callout with an embedded web view for identify results. Previously, due to the way callout determines the size of its content, the web view didn’t have an appropriate width until the map was panned enough to force the callout to refresh its layout.

This change override the web view to manually set a constant height and width. Now, when the map is tapped, the html content should be immediately visible without needing to pan the map.


nCastle1:

ThadT:
@nCastle1 I didn't see the original behavior for this (so I don't have a baseline), but it looks good to me! :)
Feel free to merge

nCastle1:
Thanks for the review! :)

PR #387, opened 2017-12-18
Removes ENC samples

nCastle1:
Sample data was not ready before release, so the samples need to be removed.


nCastle1:
Hi @dkemlage , can you please take a look at this? We need to get this in before v.next is merged into master.

nCastle1:
Thanks for the quick review! :)

PR #389, opened 2017-12-19
re-enable iOS magnifier sample

nCastle1:
This sample is not currently being shown in the sample viewer due to a groups.json omission. This PR fixes that.


nCastle1:
Note also that the code was changed to avoid a NRE; MapView interaction options is null by default.

nCastle1:
@dkemlage Can you please take a look at this? Thanks! :)

nCastle1:
Thanks for the review! :)

PR #390, opened 2017-12-19
Implement ListGeodatabaseVersions for iOS

nCastle1:
This should be the last sample implementation needed to bring iOS into
parity with all other platforms.


nCastle1:
@ThadT Can you please take a look when you have some free time?

src/iOS/Xamarin.iOS/Samples/Geoprocessing/ListGeodatabaseVersions/metadata.json
```diff
@@ -0,0 +1,14 @@
+{
+  "Name": "List geodatabase versions",
+  "SampleName": "ListGeodatabaseVersions",
+  "Description": "This sample demonstrates how to use GeoprocessingTask to get available geodatabase versions from the enterprise geodatabase. Geoprocessing task will return the versions as a table that is shown to the user in a list. This is a good example how to use geoprocessing on mapless application.",
```
ThadT: I don't like this description ... it's a little awkward. (not that I can think of anything much better right now ;)
Maybe something like:
"This sample calls a custom `GeoprocessingTask` to get a list of available versions for an enterprise geodatabase. The task returns a table of geodatabase version information, which is displayed in the app as a list."

nCastle1:
Thanks for the review! I made those changes and am now merging.

PR #391, opened 2017-12-21
Feature Layer Rendering Mode (Scene)

nCastle1:
This sample demonstrates how to control the rendering mode for a scene. It demonstrates this by showing two scene views side-by-side and allowing the user to animate a zoom between two points. The difference in rendering modes is apparent when visually comparing the scenes.


nCastle1:
@dkemlage Can you please take a look at this new sample if you have time? Thanks! :)

nCastle1:
I think there was an Android layout issue on older devices that was causing problems. I was able to simplify the layout, and with the changes, it is now working everywhere I tested. I updated the metadata.json file to include details.

PR #404, opened 2018-01-17
Enhance local server sample reliability

nCastle1:
Shutting down local server after each sample run was leading to reliability problems and was not realistic.

Now, the server will be left running. This makes starting another local server sample smoother (no popup).


nCastle1:
Hi @ThadT, can you please give this a review?

This PR corrects a problem where navigating to a sample from the tree view in WPF would cause samples to break. The samples included logic to shut down local server when the view was unshown; the 'developer use only' popup would interrupt navigation, causing the sample to be left running while Local Server had been shut down. 

This should no longer happen; although the popup can still interrupt navigation, completing the navigation works properly.

nCastle1:
Thanks for the review!

I don't think there will be any negative effects, beyond increased resource usage after starting a local server sample. Users can use the 'Local Server Services' sample to turn off local server once it has been started. Otherwise it will be shut down when the sample viewer exits.

PR #406, opened 2018-01-19
Web TiledLayer sample

nCastle1:
This sample demonstrates how to consume non-ArcGIS web tiled layers using Runtime. 

Note: the tile service the sample uses is not available over HTTPS; an Info.plist exception was already made for iOS.


nCastle1:
@dkemlage Could you please review this sample? Thanks! :)

nCastle1:
Thanks for the review! I've made that change.

PR #407, opened 2018-01-20
Format Coordinates sample

nCastle1:
This sample demonstrates how to convert between MapPoints and string coordinates. Read and written coordinate strings can be in one of several options.

Note that this PR includes some extra markdown document updates due to the nature of the automation.


nCastle1:
@ThadT Can you please give this a review? Thanks! :)

src/Android/Xamarin.Android/Samples/Geometry/FormatCoordinates/FormatCoordinates.cs
```diff
@@ -0,0 +1,201 @@
...
+
+        private void UpdateUiFromMapPoint(MapPoint startingPoint)
+        {
+            // Clear event subscriptions - prevents an infinite loop
+            _UtmEditText.TextChanged -= InputTextChanged;
+            _DmsEditText.TextChanged -= InputTextChanged;
+            _DecimalDegreesEditText.TextChanged -= InputTextChanged;
+            _UsngEditText.TextChanged -= InputTextChanged;
+
+            // Update the decimal degrees textbox
```
ThadT: Maybe change all comments that refer to "textbox" to use "text" instead (works for all platform controls).

src/Android/Xamarin.Android/Samples/Geometry/FormatCoordinates/FormatCoordinates.cs
```diff
@@ -0,0 +1,201 @@
...
+                    case "USNG":
+                        enteredPoint =
+                            CoordinateFormatter.FromUsng(_lastEdited.Text, _myMapView.SpatialReference);
+                        break;
+                }
+            }
+            catch (Exception)
+            {
+                // The coordinate is malformed, return
+                // Sample doesn't handle this because coordinates can be invalid while the user is experimenting
```
ThadT: I think this would be true if the conversion was happening on text change. Since it's kicked off from a button click, perhaps the user should be notified of invalid coordinates. Maybe a visual cue like red font color for the invalid entry?

src/Android/Xamarin.Android/Samples/Geometry/FormatCoordinates/FormatCoordinates.cs
```diff
@@ -0,0 +1,201 @@
...
+                return;
+            }
+
+            // Return if getting the point from text failed
```
ThadT: With the try/catch above, would the point ever be null? In my testing, this never happened. I think you can remove this branch (or flip it so it only calls UpdateUiFromMapPoint if enteredPoint != null.

src/Android/Xamarin.Android/Samples/Geometry/FormatCoordinates/FormatCoordinates.cs
```diff
@@ -0,0 +1,201 @@
...
+            if (enteredPoint == null)
+            {
+                return;
+            }
+
+            // Update the UI from the MapPoint
+            UpdateUiFromMapPoint(enteredPoint);
```
ThadT: Ticky-tack, I know, but both letters of a two-letter abbreviation should be capitalized: `UpdateUIFromMapPoint`. If it's three or more, only capitalize the first: `CheckHtmlResponse`, e.g. (This is the standard for .NET).

src/Android/Xamarin.Android/Samples/Geometry/FormatCoordinates/FormatCoordinates.cs
```diff
@@ -0,0 +1,201 @@
...
+            // Update the decimal degrees textbox
+            _DecimalDegreesEditText.Text =
+                CoordinateFormatter.ToLatitudeLongitude(startingPoint, LatitudeLongitudeFormat.DecimalDegrees, 4);
+
+            // Update the degrees, minutes, seconds textbox
+            _DmsEditText.Text = CoordinateFormatter.ToLatitudeLongitude(startingPoint,
+                LatitudeLongitudeFormat.DegreesMinutesSeconds, 1);
+
+            // Update the UTM textbox
+            _UtmEditText.Text = CoordinateFormatter.ToUtm(startingPoint, UtmConversionMode.LatitudeBandIndicators, true);
```
ThadT: I would suggest using `UtmConversionMode.NorthSouthIndicators` instead of `UtmConversionMode.LatitudeBandIndicators`. I think the N/S zone indicator is the more familiar way (for most folks) of viewing these coordinates.

src/Android/Xamarin.Android/Samples/Geometry/FormatCoordinates/FormatCoordinates.cs
```diff
@@ -0,0 +1,201 @@
...
+                {
+                    case "Decimal Degrees":
+                    case "Degrees, Minutes, Seconds":
+                        enteredPoint =
+                            CoordinateFormatter.FromLatitudeLongitude(_lastEdited.Text, _myMapView.SpatialReference);
+                        break;
+
+                    case "UTM":
+                        enteredPoint =
+                            CoordinateFormatter.FromUtm(_lastEdited.Text, _myMapView.SpatialReference, UtmConversionMode.LatitudeBandIndicators);
```
ThadT: I would suggest using `UtmConversionMode.NorthSouthIndicators` instead of `UtmConversionMode.LatitudeBandIndicators`. I think the N/S zone indicator is the more familiar way (for most folks) of viewing these coordinates.

src/Android/Xamarin.Android/Samples/Geometry/FormatCoordinates/FormatCoordinates.cs
```diff
@@ -0,0 +1,201 @@
...
+            _myMapView.GeoViewTapped += (sender, args) => { UpdateUiFromMapPoint(args.Location); };
+        }
+
+        private void InputTextChanged(object sender, EventArgs e)
+        {
+            // Keep track of the last edited field
+            _lastEdited = (EditText)sender;
+        }
+
+        private void ProcessTextChange(object sender, EventArgs e)
```
ThadT: Seems like the code originally tried to convert the coordinates on text change for the inputs. I think this needs a better name now that it happens as a button click: `UpdateCoordinates`, `UpdateMapPoint`, `RecalculateLocation` e.g.

src/Forms/Shared/Samples/Geometry/FormatCoordinates/FormatCoordinates.xaml
```diff
@@ -0,0 +1,32 @@
...
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="Auto" />
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="*" />
+            <ColumnDefinition Width="*" />
+        </Grid.ColumnDefinitions>
+        <Label Grid.Row="0" Grid.Column="0" Grid.ColumnSpan="2" TextColor="Red">Type in the textboxes to see a real-time conversion</Label>
```
ThadT: I think this label needs to be updated:
 - Mention that the user can click the map to see coordinates updated.
 - Remove `real-time` from the description (probably refers to the conversion that originally happened on text change?).

src/WPF/ArcGISRuntime.WPF.Samples/Samples/Geometry/FormatCoordinates/FormatCoordinates.xaml
```diff
@@ -0,0 +1,35 @@
...
+            <ColumnDefinition Width="*" />
+            <ColumnDefinition Width="*" />
+        </Grid.ColumnDefinitions>
+        <Label Grid.Row="0" Grid.Column="0" Grid.ColumnSpan="2" Foreground="Red">Type in the textboxes to see a real-time conversion</Label>
+        <Label Grid.Row="1" Grid.Column="0" HorizontalAlignment="Right">Decimal Degrees</Label>
+        <TextBox Grid.Row="1" Grid.Column="1" Tag="Decimal Degrees" x:Name="DecimalDegreesTextField" />
+        <Label Grid.Row="2" Grid.Column="0" HorizontalAlignment="Right">Degrees, Minutes, Seconds</Label>
+        <TextBox Grid.Row="2" Grid.Column="1" Tag="Degrees, Minutes, Seconds" x:Name="DmsTextField" />
+        <Label Grid.Row="3" Grid.Column="0" HorizontalAlignment="Right">UTM</Label>
+        <TextBox Grid.Row="3" Grid.Column="1" Tag="UTM" x:Name="UtmTextField" />
```
ThadT: Would be cool to implement this using data binding on the text fields and a value converter that converts the coordinate text to the correct format. (Just an observation, not a suggestion ;))

ThadT:
Note: review comments for a given platform's code often (usually?) apply to all platforms.

nCastle1:
Thanks for the review! It looks like I missed some things when I refactored halfway through 🤦‍♂️. 

I've made all of your recommended changes, except for the databinding/converter one. I'm not sure the sample content would be visible everywhere we need it to be in the SDK (e.g. in doc) given the current sampleviewer structure.

PR #408, opened 2018-01-22
New sample: StretchRaster (all .NET platforms).

dkemlage:
@nCastle1 can you please code review this new sample: StrecthRaster. 

The sample will be found in the ..\Samples\Layers\ChangeStretchRenderer hierarchy of the various SDK platforms viewers. 'StrecthRaster' is an actual type name so the filename on disk needed to be slightly different to avoid conflicts within the SampleViewer GUI framework. 

FYI....
When testing the Xamarin.Forms.Android and Xamarin.Android version of the sample, you will need to use at least the Android API 21 Emulator or an actual Android device - the KitKat (4.4) XXHDPI Phone (Android 4.4 API 19) throws an error when opening the sample . There is a Known Issue in the ArcGIS Runtime SDK for .NET at v100.1 in the Release Notes (https://developers.arcgis.com/net/latest/wpf/guide/release-notes-100-1-0.htm) that states "Xamarin Android: Some raster formats crash the app when deployed on VisualStudio Emulator" and this sample is one that exhibits this behavior.


src/Android/Xamarin.Android/Samples/Layers/ChangeStretchRenderer/ChangeStretchRenderer.cs
```diff
@@ -0,0 +1,345 @@
...
+
+            // Create the map view and add it to the main layout
+            _myMapView = new MapView(this);
+            layout.AddView(_myMapView);
+
+            // Set the layout as the sample view
+            SetContentView(layout);
+        }
+
+        private async void Initialize()
```
nCastle1: On Android and iOS we can make this `async Task`, and then make the calling method (`OnCreate` in this case) async as well. This hasn't been done in the past, but I believe it is the more correct way to do things. WPF/UWP/Forms will need to have some more work done to make this possible, so I don't recommend changes on those platforms. 

Alternatively, you could ignore the comment and the issue will be fixed consistently for all samples when it is addressed.

src/Android/Xamarin.Android/Samples/Layers/ChangeStretchRenderer/ChangeStretchRenderer.cs
```diff
@@ -0,0 +1,345 @@
...
+            // along with the parameters specified, then a stretch renderer is created and applied to 
+            // the raster layer
+
+            // Get the user choice for the raster stretch render
+            string myRendererTypeChoice;
+
+            if (_RendererTypes.SelectedItem == null)
+            {
+                // If the user does not click on a choice in the table but just clicks the
+                // button, the selected value will be null and we will use the initial
```
nCastle1: Although I personally like the 'we' language (and am probably guilty of using it in many of my own samples), I'm told this is inconsistent with the Esri technical writing style. This is such a minor nitpick that I feel bad mentioning it, but I think maybe we'll want to think about it in the future.

src/Forms/Shared/Samples/Layers/ChangeStretchRenderer/ChangeStretchRenderer.xaml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+             xmlns:esriUI="clr-namespace:Esri.ArcGISRuntime.Xamarin.Forms;assembly=Esri.ArcGISRuntime.Xamarin.Forms"
+             x:Class="ArcGISRuntimeXamarin.Samples.ChangeStretchRenderer.ChangeStretchRenderer"
+             Title="Stretch renderer">
+
+    <!--<Grid>
```
nCastle1: It looks like this was left in accidentally.

src/Android/Xamarin.Android/Samples/Layers/ChangeStretchRenderer/README.md
```diff
@@ -0,0 +1,9 @@
+# Stretch renderer
+
+This sample demonstrates how to use stretch renderer on a raster layer.
+
+<img src="ChangeStretchRenderer.jpg" width="350"/>
+
+### Instructions
+
+Choose a stretch renderer type from the list view to change the settings for the stretch renderer.\nThe sample allows you to change the stretch type and the parameters for each type. Click/tap the 'Update Renderer' button to update the raster.\nExperiment with settings for the various types for stretch parameters. For example, setting the renderer to use stretch parameters:\nMin Max with a min value of 50 and a max value of 200 will stretch between these pixel values. A higher min value will remove more of the lighter pixels values whilst a lower max will remove more of the darker.\nPercent Clip with a min value of 2 and a max value of 98 will stretch from 2% to 98% of the pixel values histogram. A lower min and higher max percentage will render using more of the original raster histogram.\nStandard Deviation with a factor of 2.0 will stretch 2 standard deviations from the mean. A higher factor (further from the mean) will render using more of the original raster histogram.
```
nCastle1: I really like the detailed explanation here :)

dkemlage:
@nCastle1 Thanks for the code review. I implemented all of your suggestion except for the modification to making the Initialize function return a Task. I figure we would fix that as a bulk fix for all of the samples in the future.

PR #409, opened 2018-01-24
Local Server deployment fixes

nCastle1:
This PR addresses two issues:
* The AGSDeployment file was improperly configured. Disabling all packages defined in the file allows all Local Server samples to run when the SDK is installed. When the local server SDK isn't installed, the sample viewer will build and run. 
* Individual samples were sometimes checking for the presence of Local Server after making calls that would cause a crash in its absence. Now, Local Server samples will not crash the sample viewer if Local Server is not installed. 
* The error message format was inconsistent between samples. Now all samples show an error message with a friendly error that directs the user to resources for installing Local Server.

### DIFF
--- a/src/WPF/ArcGISRuntime.WPF.Samples/Samples/LocalServer/LocalServerServices/LocalServerServices.xaml.cs
+++ b/src/WPF/ArcGISRuntime.WPF.Samples/Samples/LocalServer/LocalServerServices/LocalServerServices.xaml.cs
@@ -48,7 +48,7 @@ namespace ArcGISRuntime.WPF.Samples.LocalServerServices
             }
             catch (Exception ex)
             {
-                MessageBox.Show(String.Format("Please ensure that local server is installed prior to using the sample. See instructions in readme.me or metadata.json. Message: {0}", ex.Message), "Local Server failed to start");
+                MessageBox.Show(String.Format("Please ensure that local server is installed prior to using the sample. See instructions in readme.md or metadata.json. Message: {0}", ex.Message), "Local Server failed to start");
             }
         }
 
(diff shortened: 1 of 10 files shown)


nCastle1:
@dkemlage Can you please review this? You will need to uninstall the Local Server SDK to validate that the samples perform correctly without Local Server installed.

src/WPF/ArcGISRuntime.WPF.Viewer/ArcGISLocalServer.AGSDeployment
```diff
@@ -7,47 +7,47 @@ Only include these if your setup is not installing the Microsoft redistribution
   <!--The Local Server comprises a set of components which enable the use of local dynamic map and feature services based on Map Packages plus core components required for searching and querying data.
 It also provides core support for geocoding and geoprocessing.-->
   <!--Map and tile packages must be created with ArcGIS Runtime support enabled.-->
-  <Package name="Local Server" enabled="true">
+  <Package name="Local Server" enabled="false">
```
mbranscomb: @nCastle1 I recommend only setting the top level to `false`, and leaving all sub-packages set to `true`. It's then a much simpler change to get the LocalServer samples working.

src/WPF/ArcGISRuntime.WPF.Samples/Samples/LocalServer/LocalServerServices/LocalServerServices.xaml.cs
```diff
@@ -38,8 +38,18 @@ public LocalServerServices()
...
+            {
+                // Subscribe to event notification for the local server instance
+                LocalServer.Instance.StatusChanged += (o, e) =>
+                {
+                    UpdateUiWithServiceUpdate("Local Server", e.Status);
+                };
+            }
+            catch (Exception ex)
+            {
+                MessageBox.Show(String.Format("Please ensure that Local Server is installed prior to loading this sample. See the readme or metadata.json for more info. Message: {0}", ex.Message), "Local Server failed to start");
```
mbranscomb: @nCastle1 I recommend making the error message match the message strings above (`Please ensure that local server is installed prior to using the sample. See instructions in readme.me or metadata.json.`)

PR #411, opened 2018-01-25
Animate 3D Graphic

nCastle1:
This sample demonstrates how to animate a 3D model through a scene. It also demonstrates how to follow that graphic with an orbit GeoElement camera. 

Some things to note:
* I needed to add a NuGet package to get the Android UI to cooperate. It was exceedingly difficult to get the overlapping mapviews to behave without using the `ConstraintLayout`. 
* There are some variations between platforms. Some UI patterns were unworkable in a mobile form-factor, so they were left out. Generally these omissions do not substantively impact the sample and are consistent with the other platform SDKs. For example, the same sample on iOS is less featured than the sample on desktop java.


nCastle1:
@dkemlage Can you please review this?

nCastle1:
@dkemlage I believe I've addressed just about everything. Can you please re-review, with emphasis on Android?

dkemlage:
@nCastle1 Code review for the second round is complete. Everything looks great. Please Merge.
I especially like the extra code comments and sample usage instructions.
I tested that the change 'Mission' button now works on the iPad Emulator.
I tested that the Mission Progress slider now work in Xamarin.Forms.Android.
Unfortunately, the MapView inset still does not display in Xamarin.Forms.Android - oh well, not a deal breaker.
I was able to get Visual Studio to compile the Xamarin.Android version of the project - no more NuGet restore weirdness. In Xamarin.Android everything works as expected except for the MapView inset not displaying - again not a deal breaker.

Thanks,
-don

nCastle1:
@dkemlage Thanks for the review!

With regards to the Android issue, we discussed offline and believe it is a known issue involving emulators for older devices.

PR #412, opened 2018-01-26
Vector layer sample enhancements

nCastle1:
This PR includes two sets of improvements:
* The ArcGISVectorTiledLayerUrl sample now points to more interesting and up-to-date layers
* The ChangeBasemap sample now demonstrates a wider variety of basemaps, and highlights the availability of both raster and vector basemaps

Note that there are some additional UI changes for two reasons:
* The samples present many more options to the user and the previous approach didn't scale
* Our coding practices have evolved since the samples were originally made, so I chose to make them more consistent with the pattern


nCastle1:
@ThadT Sorry to ask for two very similar reviews in a row. The sample team decided that we should expand the scope of the changes after you finished the last review.

PR #416, opened 2018-02-06
Fix Sketch graphics sample on Xamarin.Forms iOS Release

nCastle1:
Testing revealed that the 'Sketch on map' sample was not working properly on Xamarin.Forms iOS when run in release mode.

That sample uses [commanding](https://blog.xamarin.com/simplifying-events-with-commanding/). In release mode, the linker was breaking the command binding as described in [this forum post](https://forums.xamarin.com/discussion/44776/command-binding-not-working-on-ios). 

This PR adjusts the linker settings in release mode to avoid the problem.


nCastle1:
@mbranscomb Can you please take a look and let me know if there are any objections? Disabling the linker feels heavy-handed, but I'm not seeing a straightforward way to prevent linking away the SketchEditor commands.

PR #419, opened 2018-02-07
Fix GeoView Sync sample

nCastle1:
This changes the GeoView Synchronization sample to also respond to navigation completed events. This will keep the views in sync even when the user flicks or uses shift+click+drag to zoom to an envelope on the map.


nCastle1:
Thanks for the review! :)

PR #423, opened 2018-02-13
Viewshed (Location) sample

nCastle1:
This sample demonstrates the configurable properties of viewshed analysis.


src/UWP/ArcGISRuntime.UWP.Samples/Samples/Analysis/ViewshedLocation/ViewshedLocation.xaml.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            // Update the viewshed settings.
+            _viewshed.Heading = HeadingSlider.Value;
+            _viewshed.Pitch = PitchSlider.Value;
+            _viewshed.HorizontalAngle = HorizontalAngleSlider.Value;
+            _viewshed.VerticalAngle = VerticalAngleSlider.Value;
+            _viewshed.MinDistance = MinimumDistanceSlider.Value;
+            _viewshed.MaxDistance = MaximumDistanceSlider.Value;
+
+            // Return if the checkboxes are in an indeterminate.
```
ThadT: " ... are in an indeterminate **state**."
Is this necessary (can it happen?). What about just `_viewshed.IsVisible = AnalysisVisibilityCheck.IsChecked == True;`?

src/WPF/ArcGISRuntime.WPF.Samples/Samples/Analysis/ViewshedLocation/ViewshedLocation.xaml.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            // Update the viewshed settings.
+            _viewshed.Heading = HeadingSlider.Value;
+            _viewshed.Pitch = PitchSlider.Value;
+            _viewshed.HorizontalAngle = HorizontalAngleSlider.Value;
+            _viewshed.VerticalAngle = VerticalAngleSlider.Value;
+            _viewshed.MinDistance = MinimumDistanceSlider.Value;
+            _viewshed.MaxDistance = MaximumDistanceSlider.Value;
+
+            // Return if the checkboxes are in an invalid state.
```
ThadT: "... **indeterminate** state."

src/UWP/ArcGISRuntime.UWP.Samples/Samples/Analysis/ViewshedLocation/ViewshedLocation.xaml.cs
nCastle1: The checkboxes use null to represent a partially checked state (checkbox is filled in rather than checked). This caught me by surprise, too. 

In the case of these samples, I don't think the check is actually needed. I get a warning from tooling about a possible `InvalidOperationException` if I remove it, but most users won't see that.

nCastle1:
Thanks for the review. I believe I've addressed everything. 

The Android deploy issue (class not found) appears to be a Visual Studio or Xamarin. Clearing the 'obj' folder in the Android project and re-building should cause the new sample to be deployed properly. Note that you may need to unload the project or close Visual Studio before the obj folder can be deleted.

PR #425, opened 2018-02-17
Download preplanned map areas

nCastle1:
This sample demonstrates how to download map areas in a preplanned workflow.


dkemlage:
@nCastle1 Thanks for the tip! Deleting the ..src\Forms\Android\obj folder and doing a re-compile fixed the Xamarin.Forms.Android problem. I can verify that the sample works on all SDk platforms.

PR #427, opened 2018-02-21
Removes reference to preview package in iOS project

nCastle1:
The iOS project never used the preview NuGet package; it appears to have been introduced to the .csproj unintentionally. Because the functionality was never referenced, the unnecessary reference never broke the build.


nCastle1:
Closing this, will be resolved as part of sample viewer upgrades.

PR #428, opened 2018-03-21
Sample Viewer Revamp

nCastle1:
This PR updates the sample viewers to use .NET attributes rather than json files to store metadata. Other changes include:

* Common implementation for sample manager, data manager, and search
* Fixes for search
* Enhanced data manager functionality
* Cross-platform 'featured' category implementation
* Annotation of sample usage of multiple class files, XAML layouts, and android axml layouts; this will be used to enhance the samples presence in the API ref documentation.


nCastle1:
@fragmer Do you have some time to review this. Apologies for the massive diff.

src/WPF/ArcGISRuntime.WPF.Viewer/Resources/ControlStyles.xaml
```diff
@@ -1,12 +1,12 @@
 ﻿<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
```
fragmer: I am not a fan of the scrollbar styling, in context. The thumb blends into the background, and the trough stands out. I tried dragging the through at first...

2018-03-23_11-00-18_arcgisruntime

src/WPF/ArcGISRuntime.WPF.Viewer/MainWindow.xaml.cs
```diff
@@ -1,100 +1,87 @@
...
-                featured.DataContext = SampleManager.Current.GetFeaturedSamples();
-                featured.SelectedIndex = 0;
+                SampleManager.Current.Initialize();
 
                 // Set category data context
-                categories.DataContext = SampleManager.Current.GetSamplesInTreeViewCategories();
+                Categories.DataContext = WPF.Viewer.Helpers.ToTreeViewItem(SampleManager.Current.FullTree);
+
+                // Select the first item
+                ((List<TreeViewItem>)Categories.DataContext).First().IsSelected = true;
```
fragmer: You can avoid the cast:

```cs
// Set category data context
var samples = WPF.Viewer.Helpers.ToTreeViewItem(SampleManager.Current.FullTree);
Categories.DataContext = samples;

// Select the first item
samples.First().IsSelected = true;
```

src/WPF/ArcGISRuntime.WPF.Viewer/MainWindow.xaml.cs
```diff
@@ -1,100 +1,87 @@
...
-            }
-        }
-
         private void categoriesList_SelectionChanged(object sender, SelectionChangedEventArgs e)
         {
             if (e.AddedItems.Count > 0)
             {
-                var sample = e.AddedItems[0] as SampleModel;
+                var sample = e.AddedItems[0] as SampleInfo;
                 SelectSample(sample);
```
fragmer: This is a non-awaited async-void call. Are you OK with the rest of this method executing before sample is done being selected?

src/WPF/ArcGISRuntime.WPF.Viewer/MainWindow.xaml.cs
```diff
@@ -103,8 +90,18 @@ private void SelectSample(SampleModel selectedSample)
...
+                    // Wait for offline data to complete
+                    await DataManager.EnsureSampleDataPresent(selectedSample);
+                }
+
+                // Show the sample
                 SampleContainer.Content = SampleManager.Current.SampleToControl(selectedSample);
-                
+
                 // Call a function to clear any existing credentials from AuthenticationManager
                 ClearCredentials();
```
fragmer: Why `GC.WaitForPendingFinalizers` in SelectSample? It's blocking.

src/ArcGISRuntime.Samples.Shared/Managers/SampleManager.cs
```diff
@@ -1,227 +1,160 @@
...
+        /// Creates a <c>SearchableTreeNode</c> representing a category of samples.
+        /// </summary>
+        /// <param name="byCategory">A grouping that associates one category title with many samples.</param>
+        /// <returns>A <c>SearchableTreeNode</c> representing a category of samples.</returns>
+        private static SearchableTreeNode BuildTreeForCategory(IGrouping<string, SampleInfo> byCategory)
+        {
+            // This code only supports one level of nesting.
+            return new SearchableTreeNode(
+                name: byCategory.Key,
+                items: byCategory.OrderBy(si => si.SampleName).ToList()
```
fragmer: This is case-sensitive. Did you meant to do `si.SampleName.ToLower()` to match the way `AllSamples` is sorted?

src/UWP/ArcGISRuntime.UWP.Viewer/Dialogs/SampleInfoDialog.xaml
```diff
@@ -15,10 +14,10 @@
             <RowDefinition Height="*"/>
             <RowDefinition Height="auto"/>
```
fragmer: I see 3 RowDefinitions for a grid with 2 controls. Something missing here?

src/UWP/ArcGISRuntime.UWP.Viewer/ErrorPage.xaml.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            InitializeComponent();
+        }
+
+        public ErrorPage(Exception ex)
+        {
+            InitializeComponent();
+
+            // TODO - do this with binding instead
+            ErrorText.Text = string.Format("{0}\n\nMessage: {1}\n\nStack Trace:\n{2}", ex.ToString(), ex.Message, ex.StackTrace);
```
fragmer: Same comment as the WPF viewer: can be reduced to `ErrorText.Text = ex.ToString()`.

src/UWP/ArcGISRuntime.UWP.Viewer/MainPage.xaml
```diff
@@ -79,8 +78,8 @@
                     <VisualState.Setters>
                         <Setter Target="ApplicationTitle.FontSize" Value="16"/>
                         <Setter Target="ApplicationTitle.Margin" Value="12,0,0,0"/>
-                        <Setter Target="SearchToggleButton.Visibility" Value="Collapsed"/>
-                        <!--<Setter Target="SearchBox.Visibility" Value="Collapsed"/>-->
+                        <Setter Target="SearchToggleButton.Visibility" Value="Visible"/>
+                        <Setter Target="SearchBox.Visibility" Value="Collapsed"/>
```
fragmer: The searchbox collapses after each keystroke when we're in the narrowest visual state. Bug?

src/UWP/ArcGISRuntime.UWP.Viewer/App.xaml.cs
```diff
@@ -39,22 +32,20 @@ public App()
...
-
-            ApplicationManager.Current.Initialize(SamplesLanguageUsed);
         }
 
         private void TaskScheduler_UnobservedTaskException(object sender, UnobservedTaskExceptionEventArgs e)
         {
             var errorMessage = e.Exception.ToString();
             Debug.WriteLine(errorMessage);
-            var message = new MessageDialog(errorMessage, "An error occurred").ShowAsync();
+            new MessageDialog(errorMessage, "An error occurred").ShowAsync();
```
fragmer: Warning CS4014 — async method call is not awaited. Consider making this handler async-void, to get rid of compiler error.

src/UWP/ArcGISRuntime.UWP.Viewer/App.xaml.cs
```diff
@@ -39,22 +32,20 @@ public App()
...
         }
 
         private void App_UnhandledException(object sender, UnhandledExceptionEventArgs e)
         {
             var errorMessage = e.Exception.ToString();
             Debug.WriteLine(errorMessage);
-            var message = new MessageDialog(errorMessage, "An error occurred").ShowAsync();
+            new MessageDialog(errorMessage, "An error occurred").ShowAsync();
```
fragmer: Warning CS4014 — async method call is not awaited. Consider making this handler async-void, to get rid of compiler error.

src/UWP/ArcGISRuntime.UWP.Viewer/Samples/Analysis/ViewshedGeoElement/ViewshedGeoElement.xaml.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        public ViewshedGeoElement()
+        {
+            InitializeComponent();
+
+            // Setup the control references and execute initialization
+            Initialize();
+        }
+
+        private async Task Initialize()
```
fragmer: Same as WPF sample — change to async-void to get rid of compiler warning.

fragmer:
I tried just about everything, but I cannot get the Android viewers to build/run. I don't have this problem with e.g. our unit-test or functional-test runners.

src/Forms/Shared/CategoryListPage.xaml.cs
```diff
@@ -1,40 +1,143 @@
...
+                // Create the sample display page to show the sample and the metadata.
+                SamplePage page = new SamplePage(sampleControl, item);
+
+                // Show the sample.
+                await Navigation.PushAsync(page, true);
+
+                // Clear Credentials
+                foreach (Credential cred in AuthenticationManager.Current.Credentials)
+                {
+                    AuthenticationManager.Current.RemoveCredential(cred);
```
fragmer: Did you mean to clear credentials _before_ or _after_ the sample is displayed? As far as I can tell, this code will run right as the sample is being displayed (i.e. as PushAsync animation is running). I'm not sure if that's intended.

src/Forms/Shared/SampleListPage.xaml.cs
```diff
@@ -1,102 +1,97 @@
...
+
+                // Create the sample display page to show the sample and the metadata.
+                SamplePage page = new SamplePage(sampleControl, item);
 
-                // Call a function to clear existing credentials
+                // Show the sample.
+                await Navigation.PushAsync(page, true);
+
+                // Call a function to clear existing credentials.
                 ClearCredentials();
```
fragmer: Same question as on CategoryListPage: what's the intended timing?

src/Forms/Shared/CategoryListPage.xaml.cs
```diff
@@ -1,40 +1,143 @@
...
+            try
+            {
+                // Load offline data before showing the sample.
+                if (item.OfflineDataItems != null)
+                {
+                    // Show the wait page.
+                    await Navigation.PushAsync(new WaitPage { Title = item.SampleName }, false);
+
+                    // Wait for the sample data download.
+                    await DataManager.EnsureSampleDataPresent(item);
```
fragmer: It's possible for user to navigate away from the WaitPage while we're awaiting this. This leads to some unexpected behavior when the rest of the method resumes. I recommend checking to see if we're still _on the same WaitPage_ before executing the rest of the method. I'd also double-check EnsureSampleDataPresent for any potential race conditions, since it's possible that it may be running multiple times concurrently. While you're at it, you can avoid flashing CategoryListPage after the download by using `Navigation.RemovePage` in place of `PopAsync`.

...Alternatively, you could use PushModalAsync/PopModalAsync to prevent user from navigating away as the wait page is displayed.

nCastle1: In testing, RemovePage (instead of PopAsync) still results in the flashed page on Android. This doesn't seem to be avoidable. I tried using modals; using remove page; only removing the page after the sample has been shown, etc.

PushModalAsync also doesn't prevent user navigation. On Android it changes the UI (hides the back button), but Android devices have a physical/virtual back button that still works.

src/Forms/Shared/CategoryListPage.xaml.cs
nCastle1: Ultimately, I don't think it matters; clearing credentials before sample load or after sample close are functionally the same (sample always runs with cleared credentials). The code could probably be clearer about the intent.

PR #432, opened 2018-03-24
Geodesic operations sample

nCastle1:
This sample demonstrates how to use geometry engine to calculate a geodesic path between two points and measure its distance.

@ThadT Do you have time to review this? Thanks!


nCastle1:
Thanks for the review! I've made changes per your comments. 

I wasn't able to reproduce the Android crash, but I suspect it has more to do with the viewer changes than it does the sample.

PR #438, opened 2018-03-29
Change bookmark in Manage Bookmarks sample

nCastle1:
Thanks to a fascinating a presentation on image classification during the DevSummit 2018 plenary, it is likely that developers will now recognize the old bookmark as being a SAM site. Hopefully the new bookmark is just as interesting.


nCastle1:
Thank you!

PR #440, opened 2018-03-30
Add search to Android sample viewer

nCastle1:
In the recent sample viewer upgrades, search was added or enhanced on all platforms except for Android. 

This PR makes two changes to the Android sample viewer:
* Implements search
* Switches to an Expandable List View (similar to TreeView) to show samples and categories on the same page. This makes the search experience a bit cleaner.


nCastle1:
@ThadT Can you please take a look at this? I can only test on device right now, so it would be great if you could test on an emulator.

ThadT:
Thanks @nCastle1, this looks great on the emulator! Feel free to merge when ready.

nCastle1:
Thanks for the review! :)

PR #442, opened 2018-04-05
Fixes UI issue in WPF viewer

nCastle1:
Previously, if you switched between samples while on the description tag, the description would be shown (with a transparent background) on top of the sample.

Now, the description will consistently have the right background.


nCastle1:
@dkemlage Can you please review this? Thanks!

PR #446, opened 2018-04-09
Licenced key

kiRyyy:
I can't find anything about how i license the map, to remove that message developers use only, can you help me? thank you :)


mbranscomb:
@kiRyyy Please see https://developers.arcgis.com/net/latest/wpf/guide/license-your-app.htm

Cheers

PR #450, opened 2018-04-18
New sample: ShowLabelsOnLayer

dkemlage:
New sample for .NET: ShowLabelsOnLayer

This is not the 1st implementation of the sample - meaning the design/layout etc. was previously dictated. The initial implementation was done on Java.


dkemlage:
@nCastle1 Can you please do the .NET code review for the various platforms (WPF, UWP, Xamarin.Forms, iOS, Android)?

PR #463, opened 2018-05-22
Update forms UI to not use tabbed page

nCastle1:
Instead, an overlay with sample details can be displayed over the sample using a frame in a relative layout. This has been tested and works well on all platforms.

Previously, the tabbed page was causing some problems with GeoViews on Android.

The image was taken from the native iOS sample viewer.


nCastle1:
@ThadT Do you mind reviewing this PR? Thanks!

PR #466, opened 2018-05-29
New sample: Densify and generalize

nCastle1:
This sample demonstrates how to use the Geometry Engine to densify and generalize (simplify) polygons.


nCastle1:
@ThadT Can you please take a look? Thanks!

src/Forms/Shared/Samples/GeometryEngine/DensifyAndGeneralize/DensifyAndGeneralize.xaml
```diff
@@ -0,0 +1,30 @@
...
+                   Text="Adjust the sliders to see the resulting generalized or densified polygon." />
+            <Label FontAttributes="Bold"
+                   Text="Max. Deviation (Generalize)" />
+            <Slider x:Name="DeviationSlider"
+                    Maximum="250" Minimum="1" Value="1" />
+            <Label FontAttributes="Bold"
+                   Text="Max. Segment length (Densify)" />
+            <Slider x:Name="SegmentLengthSlider"
+                    Maximum="500" Minimum="100" Value="100" />
+            <Label x:Name="ResultLabel"
```
ThadT: When the text in this label gets too long it wraps to a second line and pushes the stack down. This makes the UI resize and causes the map view to "twitch" a little when interacting with the slider. I think if you add `FontSize="Small"` to ResultLabel, it won't happen.

nCastle1: I retested on all platforms and no longer see this behavior with the fix applied. Thanks!

nCastle1:
@ThadT I made all requested changes. Can you please retry building on Android? I think it should be fixed, but if not we'll need to figure out what's going on. Thanks!

PR #467, opened 2018-05-30
Remove arcade expression from query.

nCastle1:
This fixes an issue with the labeling sample. 

@ThadT Can you please take a look.


ThadT:
I retested and the labels appear in Android (on my device). I'm not sure what the issue was.

PR #478, opened 2018-06-13
issues affecting all platforms - behavior - update 3 certification

ZackAllen:
Notes:
-Added more functionality to Convex Hull.
-Restructured WMTS layer sample
-Added height slider and symbol for viewpoint to Viewpoint (Location) for WPF and UWP
-Did not change code for Take Screenshot sample in Forms.
-Did not attempt to change "Show labels on layer" or "ArcGIS Tiled layer (URL)" samples

**Attempted to make the following behavior changes.**

Sample | Description | Comment / Status
-- | -- | --
Simple renderer | By default, the map zooms to the extent containing the features. Unfortunately, most of the symbols appear right at the edges of the map, making them quite hard to see. The view envelope should be a bit larger than the extent that contains the points. |  
Find place | By default, the map zooms to the extent containing the features. Unfortunately, some of the symbols appear at the edges of the map, making them quite hard to see. The view envelope should be a bit larger than the extent that contains the points. |  
Render picture markers | The campground icon is way too small. Update the sample to show the icon with an appropriate (visible to the naked eye) size |  
Find a route | The placemarks are not offset so as to be centered appropriately, so if you zoom out a lot, they move around in an unexpected way. The solution is to apply an x and y offset so that the icon is appropriately centered. |  
WMTS layer | The sample doesn't load the map by default, requiring the user to press a button. The sample should be updated so that the map loads via one of the methods at startup. |  
Surface placement | The label should be offset from the point symbol so that there is a margin between the point and the text. It looks bad currently. |  
Select ENC features | Most of the time when you tap, a large area feature will be selected. This is usually not intended and is a bad experience. Need to figure out the right way to make this not happen. |  
Convex hull list | It should be possible to reset the operation. |  
Convex Hull | The button should be disabled until a point has been placed. |  
Export tiles | The map currently centers on (0,0); it should be centered on a more interesting (and apparently deliberate) place by default. |  
Feature layer query | The sample zooms into the state too close; a padding distance should be used so that the entire state boundary can be seen when zoomed in. |  
Take screenshot | Sample code is not using an extension method properly. Also, consider freezing the image source. |  
Viewshed location | The observer is on the ground. Raise the observer a few feet to simulate a person observer |  

Screenshots will need to be replaced for a lot of these. I am holding off on taking new screenshots due to potential changes in iOS and Android Nathan warned me about. I will include updated screenshots before actually merging.


src/Android/Xamarin.Android/Samples/MapView/TakeScreenshot/TakeScreenshot.cs
```diff
@@ -54,29 +55,28 @@ private void Initialize()
             try
             {
                 // Export the image from map view
-                var exportedImage = await _myMapView.ExportImageAsync();
```
nCastle1: It looks like the extension method was used properly here, so I don't think it makes sense to change. I recommend reverting the changes to this sample for Android.

PR #479, opened 2018-06-14
Change to feature layer extrusion sample

nCastle1:
Two changes:

* The sample was using `ExtrusionMode.BaseHeight`; because there is no elevation to consider, `ExtrusionMode.AbsoluteHeight` is the better approach.
* Because of how polygons are extruded, for some polygons covering a wide area, portions of the polygon can be rendered underneath the globe surface. This PR adds a constant offset to the extrusion expression to prevent this from happening.

@ThadT Can you please take a look?

### DIFF
--- a/src/Android/Xamarin.Android/Samples/Symbology/FeatureLayerExtrusion/FeatureLayerExtrusion.cs
+++ b/src/Android/Xamarin.Android/Samples/Symbology/FeatureLayerExtrusion/FeatureLayerExtrusion.cs
@@ -74,8 +74,8 @@ namespace ArcGISRuntime.Samples.FeatureLayerExtrusion
                 // Get the scene properties from the simple renderer
                 RendererSceneProperties myRendererSceneProperties = mySimpleRenderer.SceneProperties;
 
-                // Set the extrusion mode for the scene properties to be base height
-                myRendererSceneProperties.ExtrusionMode = ExtrusionMode.BaseHeight;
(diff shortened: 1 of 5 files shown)


src/Android/Xamarin.Android/Samples/Symbology/FeatureLayerExtrusion/FeatureLayerExtrusion.cs
```diff
@@ -125,7 +125,7 @@ private void ChangeExtrusionExpression()
             // Toggle the feature layer's scene properties renderer extrusion expression and change the button text
             if (_button_ToggleExtrusionData.Text == "Population Density")
             {
-                myRendererSceneProperties.ExtrusionExpression = "[POP07_SQMI] * 5000";
+                myRendererSceneProperties.ExtrusionExpression = "[POP07_SQMI] * 5000 + 100000";
```
ThadT: Should we describe this expression (explain why we're manipulating the raw pop density, e.g.)?
Maybe something like:
`// Normalize the population density values for displaying extruded features`

 It's probably OK as is, since it's not directly relevant to what the sample is illustrating. I'll leave it up to you.

nCastle1: That's a good point, that concept is critical to explain to the user.

PR #483, opened 2018-06-19
Format Coordinates & Picture Marker fixes

ZackAllen:
Fixes in preparation for update 3. Fixed a bug in Format coordinates and fixed the size of the pin in Picture Marker.

### DIFF
--- a/src/UWP/ArcGISRuntime.UWP.Viewer/Samples/Geometry/FormatCoordinates/FormatCoordinates.xaml.cs
+++ b/src/UWP/ArcGISRuntime.UWP.Viewer/Samples/Geometry/FormatCoordinates/FormatCoordinates.xaml.cs
@@ -69,21 +69,43 @@ namespace ArcGISRuntime.UWP.Samples.FormatCoordinates
             _selectedTextField = (TextBox)sender;
         }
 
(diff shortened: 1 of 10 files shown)


src/UWP/ArcGISRuntime.UWP.Viewer/Samples/Geometry/FormatCoordinates/FormatCoordinates.xaml.cs
```diff
@@ -69,21 +69,43 @@ private void InputTextChanged(object sender, TextChangedEventArgs e)
...
 
-        private void UpdateUIFromMapPoint(MapPoint startingPoint)
+        private void UpdateUIFromMapPoint(MapPoint selectedPoint)
         {
+            try
+            {
+                // Check if the selected point can be formatted into coordinates.
+                CoordinateFormatter.ToLatitudeLongitude(selectedPoint, LatitudeLongitudeFormat.DecimalDegrees, 0);
+            }
+            catch (Exception e)
```
nCastle1: It looks like this is thrown as an `ArgumentException`. I think you could make the catch a bit more specific to just ArgumentException and then you wouldn't need an `if`. I'd also consider clearing out the textboxes and showing a mesage/alert/etc with the exception.

PR #484, opened 2018-06-20
Android sample improvements

nCastle1:
This is a focused PR that addresses the following for Android:

* Adding help labels to explain sample operation
* Fixing UI problems that interfere with sample operation
* Fixing code or behavior problems, excluding style and language feature use


src/Android/Xamarin.Android/Samples/GeometryEngine/BufferList/BufferList.cs
```diff
@@ -123,7 +123,7 @@ private void MyMapView_GeoViewTapped(object sender, GeoViewInputEventArgs e)
             {
                 // Display an error message if there is a problem generating the buffer polygon.
                 AlertDialog.Builder alertBuilder = new AlertDialog.Builder(this);
-                alertBuilder.SetTitle("Geometry Engine Failed!");
+                alertBuilder.SetTitle("Geometry Engine failed!");
```
ThadT: [sorry, I know this isn't part of the PR, feel free to ignore]
I think we should phrase this message differently and lose the exclamation (it sounds pretty harsh). It's also not necessarily accurate, I'm sure there are other reasons why this may fail. 
How about simply "Unable to generate buffers"?

src/Android/Xamarin.Android/Samples/Data/GeodatabaseTransactions/GeodatabaseTransactions.cs
```diff
@@ -339,12 +339,16 @@ private void GdbTransactionStatusChanged(object sender, TransactionStatusChanged
...
+            if (_localGeodatabase!= null && !_localGeodatabase.IsInTransaction)
             {
                 // If not, begin a transaction
                 _localGeodatabase.BeginTransaction();
                 _messageTextBlock.Text = "Transaction started";
             }
+            else
+            {
+                _messageTextBlock.Text = "Not yet ready for new transaction.";
+            }
```
ThadT: The sample seems a little flakey about initializing / getting data. It's good to have this check.

nCastle1:
@ThadT I've either made changes or filed issues for all of your suggestions, so I'll go ahead and merge. Thanks for the review! :)

PR #485, opened 2018-06-22
Updated buffer sample

ThadT:
Updated the buffer sample to show planar and geodesic buffers at a global scale.

@nCastle1 - can you please review when you have time?


src/iOS/Xamarin.iOS/Samples/GeometryEngine/Buffer/Buffer.cs
```diff
@@ -48,124 +57,245 @@ public override void ViewDidLoad()
...
+                _bufferDistanceMilesTextField,
+                _legendArea,
+                _planarSwatchLabel,
+                _geodesicSwatchLabel,
+                _clearBuffersButton);
+        }
+
+        public override void ViewDidLayoutSubviews()
+        {
+            var topMargin = NavigationController.NavigationBar.Frame.Height +
```
nCastle1: This hasn't been shared broadly with the team yet, but there is a problem where sometimes when navigating between samples, this method will be called after the view is disassociated with its controller. This causes a NullReferenceException crash. 

I recommend wrapping the body of this method in a try/catch to avoid that problem. In the future we'll find a better way (probably auto layout).

src/iOS/Xamarin.iOS/Samples/GeometryEngine/Buffer/Buffer.cs
```diff
@@ -48,124 +57,245 @@ public override void ViewDidLoad()
...
-            _bufferDistanceMilesUITextField.Text = "10";
-            _bufferDistanceMilesUITextField.AdjustsFontSizeToFitWidth = true;
-            _bufferDistanceMilesUITextField.BackgroundColor = UIColor.White;
-            // - Allow pressing 'return' to dismiss the keyboard
-            _bufferDistanceMilesUITextField.ShouldReturn += (textField) => { textField.ResignFirstResponder(); return true; };
-
-            // Add the MapView and other controls to the page.
-            View.AddSubviews(_myMapView, _bufferInstructionsUILabel, _bufferDistanceMilesUITextField);
+            // Create the help label.
+            _helpLabel = new UILabel
```
nCastle1: I recommend putting a UIToolbar between the help label as well, and also updating the toolbar at the bottom to be behind both rows of controls.

src/Forms/Shared/Samples/GeometryEngine/Buffer/Buffer.xaml
```diff
@@ -2,21 +2,56 @@
...
-                     Orientation="Vertical">
-            <Label x:Name="BufferInstructionsLabel" 
-                   Text="Specify the buffer in miles and the tap on the map:"/>
-            <Entry x:Name="BufferDistanceMilesEntry" 
-                   Text="10"/>
-        </StackLayout>
-        <esriUI:MapView x:Name="MyMapView" 
-                        Grid.Row="1"/>
+        <esriUI:MapView x:Name="MyMapView"/>
+        <Frame BackgroundColor="DarkGray" Opacity="0.8"
```
nCastle1: Although this follows the pattern on UWP and WPF, it can be difficult to use on the phone (especially simulators because of how they implement multitouch). 

I recommend moving this above the mapview so that they don't overlap. It will be a bit uglier on most platforms, but should be easier to use.

src/Android/Xamarin.Android/Samples/GeometryEngine/Buffer/Buffer.cs
```diff
@@ -15,135 +15,177 @@
...
+                // Get the location tapped by the user (a map point in the WebMercator projected coordinate system).
+                MapPoint userTapPoint = e.Location;
 
-                // Get the buffer size from the UITextField.
+                // Get the buffer distance (miles) entered in the text box.
                 double bufferInMiles = System.Convert.ToDouble(_bufferDistanceMilesEditText.Text);
 
-                // Create a variable to be the buffer size in meters. There are 1609.34 meters in one mile.
+                // Convert the input distance to meters. There are 1609.34 meters in one mile.
                 double bufferInMeters = bufferInMiles * 1609.34;
```
fragmer: Tip: We've got a helper API for this in Runtime:
```cs
double bufferInMeters = LinearUnits.Miles.ToMeters(bufferInMiles);
```

ThadT:
Thanks @fragmer and @nCastle1 
I've incorporated most of the feedback. There are some definite improvements needed for the iOS and Android UIs, but I think the sample is good for now.

PR #486, opened 2018-06-25
Changed "FeatureCollectionLayerFromPortal" to use a different portal item by default

ZackAllen:
Also added punctuation to comments and updated the year.


ThadT:
@ZackAllen I think you can go ahead and merge this now.

PR #500, opened 2018-07-11
Source code tab - WPF & UWP

ZackAllen:
Added a source code viewer tab for both of these platforms.  Syntax highlighting of source code is done using the same .js and .css as the doc website. Works for all samples. Can work with multiple .xaml and .cs files.

This brings significant changes to the .csproj files. Additional lines are necessary for every .cs and .xaml.cs file in UWP. Additional lines are necessary for every .xaml file in WPF.

I am aware there are some daily build issues with the .csproj files. I will fix these when merging.


src/UWP/ArcGISRuntime.UWP.Viewer/ArcGISRuntime.UWP.Viewer.csproj
```diff
@@ -1250,14 +1268,16 @@
       <Version>100.3.0</Version>
     </PackageReference>
     <PackageReference Include="Esri.ArcGISRuntime.UWP">
-      <Version>100.3.0</Version>
+      <Version>100.3.0-daily2102</Version>
```
nCastle1: We need to remove references to the daily builds in both this csproj and the WPF version.

src/UWP/ArcGISRuntime.UWP.Viewer/ArcGISRuntime.UWP.Viewer.csproj
```diff
@@ -111,8 +111,19 @@
...
+  
   <!-- Screenshots -->
   <ItemGroup>
+    <Content Include="Resources\SyntaxHighlighting\highlight.css">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Resources\SyntaxHighlighting\highlight.pack.js">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+    <Content Include="Resources\SyntaxHighlighting\screen.css">
```
nCastle1: It looks like you can do without `screen.css`. The behavior I'm seeing is that it adds a margin at the top of the page and makes it so you have to scroll all the way down to see the horizontal scroll bar. I think the styles there make more sense in the context of showing a code block on a web page than they do when showing a full-page source browser.

src/UWP/ArcGISRuntime.UWP.Viewer/SourceCodeViewer.xaml
```diff
@@ -0,0 +1,28 @@
...
+            <RowDefinition Height="auto" />
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="*" />
+        </Grid.ColumnDefinitions>
+        <ComboBox x:Name="FileSelection"
+                          Width="auto"
+                          Grid.Row="0"
+                          HorizontalAlignment="Right"
```
nCastle1: I recommend setting this value to 'stretch' to fill the width.

src/UWP/ArcGISRuntime.UWP.Viewer/SourceCodeViewer.xaml
```diff
@@ -0,0 +1,28 @@
...
+             mc:Ignorable="d"
+             Background="{StaticResource PrimaryBackgroundColor}">
+
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="auto" />
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="*" />
```
nCastle1: I recommend reviewing this file for opportunities to simplify. I believe the column definitions can be removed, as well as setting width and height to auto on the ComboBox and WebView. There may be other simplifications that can be made.

src/UWP/ArcGISRuntime.UWP.Viewer/SourceCodeViewer.xaml
```diff
@@ -0,0 +1,28 @@
+﻿<UserControl x:Class="ArcGISRuntime.UWP.Viewer.SourceCode"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:models="clr-namespace:ArcGISRuntime.Samples.Shared.Models"
```
nCastle1: I don't think this is needed.

src/WPF/ArcGISRuntime.WPF.Viewer/Resources/SyntaxHighlighting/highlight.css
```diff
@@ -0,0 +1,104 @@
+.xml .hljs-tag .hljs-attribute, .html .hljs-tag .hljs-attribute {
+    color: #aa0d91;
+}
+/*
+ XCode style (c) Angel Garcia <angelgarcia.mail@gmail.com>
```
nCastle1: I recommend using a theme that is similar to the one used by Visual Studio, because that will be most comfortable/familiar for our developers. It looks like there is one available [here](https://highlightjs.org/static/demo/styles/vs.css)

ZackAllen: I am currently using the one from the documentation website. Purpose is to have consistency with the website. What do you think is better? I tired your css that you linked and it looks nearly identical.

nCastle1: The differences are very minor and given that it is consistent with the website, I think that's totally fine.

PR #512, opened 2018-08-01
Android Forms fix for OAuth sample

ThadT:
Added static property to the MainActivity class to provide the app context.


ThadT:
Hey @nCastle1 
I implemented a fix for the problem reported in the OAuth sample. I only updated a few lines of code, let me know if this looks OK.

PR #513, opened 2018-08-01
Remove the custom page renderers from the Xamarin Forms projects.

ThadT:
The Xamarin Forms Android and iOS projects originally used page renderers to implement platform-specific OAuth for the "Author map" sample. The logic for that sample has been reworked to use conditional code blocks instead.
@nCastle1 - Another small (low priority) PR, please take a look when you have a chance.


ThadT:
Thanks @nCastle1, good catch. I didn't realize the page renderer was in play anywhere else in the app.

src/Forms/Shared/Samples/Security/OAuth/OAuth.xaml.cs
```diff
@@ -151,7 +151,7 @@ public async Task<Credential> CreateCredentialAsync(CredentialRequestInfo info)
 
 #if __ANDROID__
             // Get the current Android Activity
-            Activity activity = (Activity)Android.App.Application.Context;
```
nCastle1: Was this intentional? I think this is competing with other changes.

nCastle1: Note: I remembered that there's another PR open that addresses this.

PR #517, opened 2018-08-20
Readme display improvements

nCastle1:
This PR updates the sample viewers to display formatted readmes from the readme files for each sample. 

Note: many readmes needed to be renamed due to case-sensitive filesystems on iOS and Android.


nCastle1:
@ThadT I only made the change for the viewers that already had readme-viewing capabilities. There is an issue to implement for native iOS and Android later.

PR #519, opened 2018-08-22
Update WPF samples for Local Server 100.3

nCastle1:


src/WPF/ArcGISRuntime.WPF.Viewer/ArcGISLocalServer.AGSDeployment
```diff
@@ -1,109 +0,0 @@
...
-  </Package>
-  <!--Enables debugging options within the runtime.-->
-  <!--Debugging options should only be used for testing purposes, and not deployed in final solutions.-->
-  <Package name="Debug" enabled="false">
-    <ChildPackages>
-      <!--Adds support for runtime logging.-->
-      <Package name="Logging" enabled="false" />
-    </ChildPackages>
-  </Package>
-</Packages>
```
mbranscomb: Keeping this deployment file would allow for switching between Local Server v100.3 and Local Server v100.2. Unlikely to be a requirement in the context of the samples, but worth considering before removing.

PR #520, opened 2018-08-24
Updates to WMS service catalog

nCastle1:
This PR makes the following changes to the WMS service catalog sample across all platforms:
* Updated readme content with increased detail
* Updated the layer list to display as a tree (either simulated with spaces or with a TreeView control if available)


ThadT:
Awesome, looks great. Feel free to merge.

PR #526, opened 2018-09-15
Sample cartography updates

nCastle1:
As part of an effort to make our samples more interesting and reliable, we've updated the services & symbology used by several samples.


nCastle1:
@ThadT Can you please review when you get a chance?

nCastle1:
Issue with open scene sample is fixed. I can't reproduce the Forms readme issue - let's meet up to investigate that. In the meantime, I'm merging this PR; we can make another to fix the issue.

PR #535, opened 2018-09-28
Fixes for Android configuration in release & non-https URL in sample

nCastle1:
Some issues causing http problems on Android discovered during certification:

* Mono HttpClient implementation causes problems in release, so I switched to the native implementaiton
* Load web tiled layer was referring to stamen tiles over http; I updated to use https so that the sample will work on modern Android, which is adopting an iOS-like security policy


nCastle1:
@ThadT I believe that's expected. It would be really nice if we could put real links in the attribution bar!

PR #539, opened 2018-10-07
Fix Android AXML namespaces

nCastle1:
There is a change to the API that moves the UI controls from `Esri.ArcGISRuntime.UI.` to `Esri.ArcGISRuntime.UI.Controls`. At run time, this causes an exception when loading Android layouts with GeoViews. This PR fixes the samples so that they work with 100.4.


nCastle1:
Thanks!

PR #548, opened 2018-10-12
Fix issue with Android slider min value

nCastle1:
Android progress bars don't support setting a minimum value on all of our supported SDKs. This sample removes references to ProgressBar.Min.


nCastle1:
@dkemlage Can you please review and approve? Thanks!

pragatisahu:
Works as expected in Lollipop and Nougat android version.

PR #562, opened 2018-11-12
Improve handling of exceptions raised from async void methods

nCastle1:
This PR updates several samples to improve handling of exceptions that may be thrown by Runtime from async void methods. 

This PR has been tested on all platforms.


nCastle1:
@dkemlage Can you please review? Thanks!

PR #563, opened 2018-11-13
New sample: identify layers

nCastle1:
This sample demonstrates how to use MapView to identify features within multiple layers.


dkemlage:
@nCastle1 Approved for merging.
I tested all platforms and the sample works without issue.

PR #564, opened 2018-11-16
Fix UWP build issue

nCastle1:
In a previous PR, a build-breaking error was (not) discovered. This PR fixes it for UWP and Xamarin.Forms UWP.


nCastle1:
Thanks for the review!

PR #565, opened 2018-11-21
Misc. Forms fixes

nCastle1:
Several issues found during 100.4 certification are fixed by this PR. See individual commits for details about changes.

Note: also adds support for new iPhones. This didn't come up during certification, but it was looking really bad on some devices.


nCastle1:
@dkemlage Can you please review this? I've tested and everything works well on iOS, Android, and UWP. This affects Forms samples only.

PR #573, opened 2018-12-14
Sample UI improvements for iPhone X

nCastle1:
PR is WIP

### DIFF
--- a/src/iOS/Xamarin.iOS/Samples/Data/ListRelatedFeatures/ListRelatedFeatures.cs
+++ b/src/iOS/Xamarin.iOS/Samples/Data/ListRelatedFeatures/ListRelatedFeatures.cs
@@ -30,8 +30,8 @@ namespace ArcGISRuntime.Samples.ListRelatedFeatures
     public class ListRelatedFeatures : UIViewController
     {
         // Create and hold references to the UI controls.
-        private readonly MapView _myMapView = new MapView();
-        private UITableView _myDisplayList = new UITableView();
+        private MapView _myMapView;
+        private UITableView _tableView;
 
         // URL to the web map.
         private readonly Uri _mapUri =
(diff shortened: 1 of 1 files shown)


src/iOS/Xamarin.iOS/Samples/Data/StatsQueryGroupAndSort/StatsQueryGroupAndSort.cs
```diff
@@ -748,40 +728,20 @@ public class ChooseStatisticOverlay : UIView
...
-            };
-            toolbar.SizeToFit();
+            UIToolbar toolbar = new UIToolbar();
+            toolbar.TranslatesAutoresizingMaskIntoConstraints = false;
+
+            statPicker.TranslatesAutoresizingMaskIntoConstraints = false;
 
             // Add Button (add the new stat and don't dismiss the UI).
-            UIBarButtonItem addButton = new UIBarButtonItem("Add", UIBarButtonItemStyle.Done, (s, e) =>
+            UIBarButtonItem addButton = new UIBarButtonItem("Add", UIBarButtonItemStyle.Plain, (s, e) =>
```
fragmer: Almost forgot one more:
Clicking "Add" button on "Stat definitions" screen does nothing unless user changes something in the picker first.

nCastle1:
@dkemlage Thanks for the review! I've fixed the two crashes you found, which were indeed a side effect of the changes I made. 

@fragmer Thanks for the review! I've fixed all but the last two comments. There's some trickiness with keyboard types that I'd need to account for (in particular, the lack of a return key). Combined with the complexity of implementing given the current structure, it might be better to wait until the sample is redesigned. For stats query, there would need to be significant restructuring to avoid the issue you ran into.

PR #577, opened 2018-12-24
Change atmosphere effect

nCastle1:
This PR implements change atmosphere effect for all platforms.


nCastle1:
@dkemlage I've made all suggested changes. Please re-review when you can. Thank you!

PR #582, opened 2019-01-11
Set the AppDataPath for local server in all LS samples

nCastle1:
This is a necessary step to avoid complications when `%LOCALAPPDATA%` is very long.


mbranscomb:
@nCastle1 That approach looks good. The `ArcGISRuntimeEnvironment.TempPath` Property should also be set. Perhaps consider:
- `AppDataPath` = `Root + EsriSamples\AppData`
- `TempPath` = `Root + EsriSamples\Temp`

PR #587, opened 2019-01-17
New sample: manage operational layers

nCastle1:
This sample demonstrates how to add, remove, and reorder layers in a map. 

Note: this sample has a rich UI that required some variation between platforms. Some of the code is significantly different between platforms as a result.


nCastle1:
@dkemlage Thanks for the review!

PR #589, opened 2019-01-18
Merges master into v.next

nCastle1:
The recent iOS PR was against master, but all new work is now going into v.next. 

This PR updates v.next with the latest content from master.


nCastle1:
Thank you!

PR #591, opened 2019-01-28
New sample: OpenStreetMap layer

nCastle1:
This PR implements the OpenStreetMap layer sample for all platforms.


nCastle1:
@pragatisahu I was able to fix the link issue with the webview on Forms.

PR #592, opened 2019-01-28
New sample: Project

nCastle1:
This sample demonstrates how to project a point from one spatial reference to another.


nCastle1:
Callouts on iOS don't respect the newline character, so it isn't possible to achieve the desired formatting without adding a lot of extra UI code. Given that this is one of the more focused samples I think bolding the projected coordinates is a reasonable compromise in the name of simplicity.

PR #593, opened 2019-01-29
New sample: reverse geocode

nCastle1:
This sample demonstrates how to find the address for a tapped point using the world geocoder service.


nCastle1:
First two changes are made.

I don't see a way to make the callout text wider or to wrap. As an alternative, the geocode result could be displayed in a textview above the map, but I don't know if it would be as easy to notice that there was a result. Another option is to show an alert with the geocoded address.

PR #594, opened 2019-01-29
New sample: offline geocode

nCastle1:
This sample demonstrates how to perform forward and reverse geocoding using an offline locator.


pragatisahu:
Verified the sample with all the changes. Looks good to me.

PR #595, opened 2019-01-30
New sample: Mobile map search and route

nCastle1:
This sample demonstrates how to open a mobile map package and use available networks, locators, and maps.


nCastle1:
It looks like the readme was changed between the review being done and the review being posted, because I can't find the typo mentioned.

I'm not sure there is a tolerance issue. The locator has a more detailed view of the transportation network than is shown in the map, and it will also find the nearest address for locations on the water. I don't think this behavior can be changed.

pragatisahu:
Verified the changes. Sample looks good to me.

PR #610, opened 2019-03-08
Fix UWP & Android UI issues

nCastle1:
This fixes an issue with the terrain elevation slider in UWP. Previously it was snapping to integer values, where other platforms allowed more fine-grained adjustments. 

Also updates the Android samples to not automatically reload when the device is rotated.


nCastle1:
It looks like per-app rotation control is a new feature in Android Pie. https://www.howtogeek.com/fyi/google-hid-an-awesome-per-app-auto-rotate-feature-in-android-p/

PR #613, opened 2019-03-12
New sample: Display KML tour

nCastle1:
Demonstrates how to consume KML tour content in ArcGIS Runtime.


nCastle1:
1. The audio should be in sync. I believe there was a recently fixed issue
2. Yes, the placemark should float for those locations.
3. Yes, that workflow is expected. The API provides no way to know when the tour ends.

PR #618, opened 2019-03-18
Update KML tour sample to use PCE

nCastle1:
I had missed the KmlTour.TourStatus property. Using PropertyChanged event, it is possible to detect changes to the tour status, which enables updating the UI when the tour completes.

Two other changes:
* Changed some code to use a new pattern `if (node is KmlContainer container)` which saves a line and hopefully makes the intent clearer.
* Added an Android package reference. I believe this should have done in a previous PR, but possibly due to some caching/partial build issue I didn't see the error previously.


nCastle1:
@pragatisahu I've updated the samples to fix some of these issues. Can you please re-review? Thanks!

pragatisahu:
Verified with build 2366. 
Tested issue 2 and 3 and they worked perfectly fine now. 
I believe issue 1 has a core-api issue.
Ran into no other issues. The sample looks good to me.

PR #620, opened 2019-03-21
New sample: Group layers

nCastle1:
This sample demonstrates the new group layer support introduced at 100.5.


nCastle1:
@pragatisahu I've updated the ordering of the layers to be consistent between platforms. Re: sublayers still being selected even when invisible, that is part of the design of the sample. I think it is useful in highlighting how the API behaves, especially since it isn't necessarily intuitive compared to how other GIS applications work.

pragatisahu:
@nCastle1 : Thank you for letting me know that.
I verified the sample with build 2366. The sample looks good to me.

PR #621, opened 2019-03-25
Update terrain exaggeration minimums

nCastle1:
There was a change to the design of the sample. The new minimum value for the terrain exaggeration slider is `1` at v100.5.


pragatisahu:
Verified with build 2373.
Ran into no issue.
(Can no longer reproduce the starfield through atmospheric effect issue)

PR #627, opened 2019-04-09
Fix iOS memory issues by following viewcontroller lifecycle best practice

nCastle1:
Previously, the iOS samples weren't using the viewcotnroller lifecycle methods to subscribe and unsubscribe from events. This caused a memory leak due to how the garbage collector is implemented on Xamarin.iOS. 

The memory leak issue would cause a crash if several samples were used in sequence, because iOS kills apps that use too much memory. 

Changes:

* `UIBarButtonItem`s are held in fields
* Event subscriptions added in `ViewWillAppear`
* Event subscriptions removed in `ViewDidDisappear`
* TableView references to the owning ViewController are marked `[Weak]`
* Jobs and timers are canceled in `ViewDidDisappear`
* Comments have been standardized in a few relevant areas, mainly by the UI field declarations, in `ViewWillAppear` and `ViewDidDisappear`, and in `LoadView`. 

How to test:

I've already tested all samples to confirm that the memory leak issues are fixed. This is a large change with relatively high risk of error, especially when managing event re-subscriptions (sample startup). 

To test, all samples should be opened and closed. If there are any leaks, the app will be killed before all samples can be run. 

Note: the following samples have lingering memory issues, primarily due to a difficult-to-debug issue with Xamarin.Auth:

* Author map
* Search a portal for maps
* Generate an offline map
* Generate an offline map with overrides
* Integrated windows authentication
* OAuth
* Token secured challenge
* Stats query group and sort

Those samples should be excluded from the test procedure defined above.


pragatisahu:
Verified with build: 2381 and 2385.
* Random crashes: 1. (was going through the samples under Featured category and the sample crashed at Raster hillshade renderer) 
* Mobile map (Search and Route): crash while downloading the data.
* Open mobile scene packages: fails to download data and threw the below error.
[error](https://user-images.githubusercontent.com/40747019/55991185-7f10cf00-5c5e-11e9-9a82-86bdc5f8c7d7.png)
* WMS service catalog: not showing up timestamp data.
* Feature layer dictionary renderer: failed to load data and threw the below error.
[error](https://user-images.githubusercontent.com/40747019/55991404-0fe7aa80-5c5f-11e9-8885-a1396cf9e680.png)

Overall I felt the downloading of data for samples took longer time than usual.

nCastle1:
Re: version issue, I think we need to expect users to be on the latest, and to have the same version on both platforms. The 'Link Framework SDKs only' option works in this specific scenario, but users would run into issues if they tried accessing newer SDKs than what they have available.

PR #628, opened 2019-04-09
Fixes for issues found in certification

nCastle1:
Issues addressed:

* **iOS** - Convex hull wasn't resetting old hull before showing new one.
* **All platforms** - geopackage samples need to be updated for new path behavior
* **Android samples** - alert dialog text unreadable after theme change
* **Android samples** - some samples were missed in earlier effort to prevent sample reset on rotation
* **Android** IWA - keyboard autocompletes in URL field
* **Forms Android** IWA - once the list of maps is scrolled out of view, it couldn't be scrolled back in
* **Forms Android** Find Place - selecting a suggestion result for a location updated the search instead
* **All platforms** - Service area interactive didn't allow finishing drawing barriers
* **UWP** - Find address - selecting a suggestion didn't always zoom to that address.
* **Forms iOS** - Author and save map - couldn't add more than one layer to map
* **WPF and UWP** Author and save a map - sample title updated to be consistent with other platforms


pragatisahu:
Verified with build 2390. Ran into no issues, all changes looks good to me.

PR #631, opened 2019-04-13
An example of how to do some telemetry in the WPF Sample app using AppInsights.

dotMorten:
This tracks all samples that gets loaded in what order, and a few key system metrics (like graphics card/driver, OS/Framework versions, remote desktop etc) that have been known to cause various rendering problems.

To enable this, declare an msbuild variable `AppInsightsKey` that matches the app insights app center key, and either manually update the key in App.config, or use a build-script in the CI build to do it.
Without any of this, the user won't see any difference to the app (we don't want telemetry from locally modified apps, as this could provide false positives)

For instance in Azure Pipelines, it can be done like this: https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/blob/3e083f4e541dc71ee9a4fe68c492ef3af07dc52b/azure-pipelines.yml#L73-L76
The variable is then secretly declared in the Azure Pipeline and stays internal.

image

Thanks to @onovotny for the steps to do this.

Note: There's some definite to-do's here, most importantly we need a way for users to opt in/out of this in the settings page, the first run of the app should ask users for permission, and a privacy policy should be included. Therefore this is more a draft/proof-of-concept.

### DIFF
--- a/src/WPF/ArcGISRuntime.WPF.Viewer/App.config
+++ b/src/WPF/ArcGISRuntime.WPF.Viewer/App.config
@@ -3,6 +3,9 @@
     <startup> 
         <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
     </startup>
+  <appSettings>    
+    <add key="InstrumentationKey" value="__AppInsightsKey__" />
+  </appSettings>
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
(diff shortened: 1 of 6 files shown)


onovotny:
Keep in mind, that in a client app, the app insights key isn't really a secret. The client can see it by looking at the config. It's the same with web apps as well if you do javascript instrumentation (which is the default), as the instrumentation key is visible there too. 

I do it in Pipelines mostly as a way to keep it out of source control.

dotMorten:
@onovotno thanks I'm aware. It's for the same reason I'd like to do it.

ZackAllen:
This is no longer needed now that Microsoft App Center analytics have been implemented.

PR #635, opened 2019-04-26
fixed bug caused by duplicate rasterlayer names

kirkktx:
I was getting an exception, this fixes it


nCastle1:
Hi Kirk,

Thank you for opening this PR to fix the sample. 

As part of choosing a fix for the issue, the samples team decided to simplify the sample to focus more on the core functionality of opening the GeoPackage. That change was included as part of #628, which will be merged within the next few days. 

My apologies for taking so long to respond; I've updated my GitHub notification settings so I should be more responsive going forward.

PR #673, opened 2019-08-20
Minor UWP enhancements

nCastle1:
* Append platform to app title: We have three apps installed as Windows packages: WPF viewer for Store, UWP, and Forms UWP. This PR updates the app manifests to include a label for each to make it clear which one you're looking at.
* Collapse tree view when canceling search - matches WPF viewer behavior

### DIFF
--- a/src/Forms/UWP/Package.appxmanifest
+++ b/src/Forms/UWP/Package.appxmanifest
@@ -15,7 +15,7 @@
   </Resources>
   <Applications>
     <Application Id="App" Executable="$targetnametoken$.exe" EntryPoint="FPCL.WIndows.App">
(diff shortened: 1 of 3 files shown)


mbranscomb:
@nCastle1 I verified the addition of `(UWP)` to the native UWP viewer - provides good clarification.

When running the UWP viewer I received an...

```
Unhandled exception at 0x00007FFA3E5426F0 (Windows.UI.Xaml.dll) in ArcGISRuntime.exe: 0xC000027B: An application-internal exception has occurred (parameters: 0x00000246ACFFDB10, 0x0000000000000002). occurred
```

PR #676, opened 2019-08-21
Update TOC markdown format for links

nCastle1:
Per the markdown spec, URLs can't have spaces in them. This PR fixes both the markdown TOC files and the script that generates them.


jf990:
i tested a handful of the links on your branch they render and link correct. thanks!

PR #697, opened 2019-09-24
iOS 13 Fixes

ZackAllen:
Updated Xamarin.iOS viewer and Xamarin.Forms iOS viewer to be compatible with iOS 13.


saratchandrakarumuri:
Compiles well with latest Xcode and working on iOS 13 and 11 devices as expected

PR #716, opened 2019-11-25
100.7 Sample update: Surface placements

ZackAllen:
Sample now inclues draped flat and draped billboard. The user can switch between the two placement modes using UI controls.


src/iOS/Xamarin.iOS/Samples/GraphicsOverlay/SurfacePlacements/SurfacePlacements.cs
```diff
@@ -23,12 +23,17 @@ namespace ArcGISRuntime.Samples.SurfacePlacements
...
         "Surface placement",
         "GraphicsOverlay",
-        "This sample demonstrates how to position graphics using different Surface Placements.",
+        "Position graphics relative to a surface using different surface placement modes.",
         "")]
     public class SurfacePlacements : UIViewController
     {
         // Hold references to UI controls.
         private SceneView _mySceneView;
+        private UISegmentedControl _picker;
```
nCastle1: I recommend putting the segmented control in a toolbar; this will look better on iPhone and be more appropriate on larger devices like iPad.

PR #718, opened 2019-11-27
100.7 Sample: Display annotation

ZackAllen:
New sample showing an annotation layer.


src/Android/Xamarin.Android/Samples/Layers/DisplayAnnotation/DisplayAnnotation.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace ArcGISRuntimeXamarin.Samples.DisplayAnnotation
+{
+    [Activity(ConfigurationChanges = Android.Content.PM.ConfigChanges.Orientation | Android.Content.PM.ConfigChanges.ScreenSize)]
+    [ArcGISRuntime.Samples.Shared.Attributes.Sample(
+        "Display annotation",
+        "Layers",
+        "Display annotation from a feature service URL.",
+        "")]
+    [ArcGISRuntime.Samples.Shared.Attributes.OfflineData()]
```
nCastle1: I recommend removing this attribute - I think its an issue with the sample generator script that includes it even when you don't specify any offline data items

PR #719, opened 2019-12-02
Forms.iOS: Fix for iOS devices with curved bottoms

ZackAllen:
This is a UI fix for xamarin forms iOS. It adds padding to the bottom to prevent the map from overlapping with the bottom control on iOS devices. 34 is not a magic number, it is the actual safe value for from the bottom control.

### DIFF
--- a/src/Forms/Shared/SamplePage.xaml
+++ b/src/Forms/Shared/SamplePage.xaml
@@ -1,29 +1,49 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
-<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
-            xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-            x:Class="ArcGISRuntime.SamplePage"
-            >
+<ContentPage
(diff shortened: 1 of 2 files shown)


src/Forms/Shared/SamplePage.xaml
```diff
@@ -1,29 +1,49 @@
...
-               RelativeLayout.XConstraint="{ConstraintExpression Type=RelativeToParent,Factor=.05,Property=Width}"
-               RelativeLayout.HeightConstraint="{ConstraintExpression Type=RelativeToParent, Factor=.9,Property=Height}"
-               RelativeLayout.YConstraint="{ConstraintExpression Type=RelativeToParent, Factor=.05,Property=Height}"
-               Padding="0"
-               IsVisible="False"
-               CornerRadius="0">
-            <WebView x:Name="DescriptionView"/>
+        <Frame
+            x:Name="SampleContentPage"
+            Padding="0"
```
nCastle1: I recommend setting `HasShadow` to false to avoid emphasizing the cut out area.

PR #729, opened 2019-12-16
Browse WFS Layers: Design update

ZackAllen:
Updated the design for Browse WFS layers to omit coordinate swapping, as data used in the sample doesn't need swapping. This matches other platforms.

Also replace the random colors with blue, to guarantee visibility of features against the basemap.


ZackAllen:
Will not make this change until design is finalized.

PR #730, opened 2019-12-17
100.7 Android Certification Fixes

ZackAllen:


cgutierrez-gis:
All fixes looking good aside from the completed load dialog not dismissing once the map has been taken offline (i.e once it has reached 100%) on the Generate Offline Map (overrides) sample. Once there is some interaction with the screen though the alert is dismissed and the fully offline map is presented.

PR #733, opened 2019-12-18
Simple Renderer implementation update

ZackAllen:
The Add graphics (SimpleRenderer) sample currently has an inconsistent bug as forms. the `Initialize` method has a race condition with the mapview loading that can put the yellow points outside of the starting viewpoint. I reworked the sample to eliminate this and have a better workflow in general. The corner points are now created from an envelope instead of doing math on a center point.


cgutierrez-gis:
UWP issue corrected by placing yellow points within the correct zoomed viewpoint, no negative impact on sample in Android & iOS from changes.

PR #761, opened 2020-03-04
Add all changes in master to v.next

ZackAllen:
* Adds several fixes that were only in master
* Moves configure subnetwork trace to utility network category

I've built this successfully on every platform.


src/Android/Xamarin.Android/Samples/Security/OAuth/readme.metadata.json
```diff
@@ -19,9 +23,12 @@
         "/net/latest/android/sample-code/oauth.htm"
     ],
     "relevant_apis": [
-        "ArcGISPortal",
-        "AuthenticationManager",
-        "Credential"
+        "",
```
nCastle1: This looks like an issue - might need to be fixed on both branches

ZackAllen: Was due to an errant space in the readmes. I've fixed in this branch and will do a separate PR for master

PR #783, opened 2020-04-29
100.8.0 Cert fixes: Xamarin.iOS

ZackAllen:


src/iOS/Xamarin.iOS/Samples/Analysis/DistanceMeasurement/DistanceMeasurement.cs
```diff
@@ -48,6 +48,8 @@ public DistanceMeasurement()
 
         private void Initialize()
         {
+            _mySceneView.GeoViewTapped -= MySceneView_GeoViewTapped;
```
nCastle1, line 51: Initialize is consistently called after `ViewWillAppear`, so I don't think this is necessary. I would recommend removing this, and also removing the event subscription at the end of `Initialize`. I recommend adding a comment in place of the event subscription at the end of `Initialize` indicating that the event subscription happens in `ViewWillAppear`. 

I also recommend replacing the measurement changed handler lambda with a method and subscribing/unsubscribing in `ViewWillAppear`/`ViewDidDisappear`.

src/iOS/Xamarin.iOS/Samples/Geometry/BufferList/BufferList.cs
```diff
@@ -316,7 +316,7 @@ public override void LoadView()
             _bufferDistanceEntry.Layer.CornerRadius = 4;
             _bufferDistanceEntry.LeftView = new UIView(new CGRect(0, 0, 5, 20));
             _bufferDistanceEntry.LeftViewMode = UITextFieldViewMode.Always;
-            _bufferDistanceEntry.KeyboardType = UIKeyboardType.NumberPad;
```
nCastle1, line 319: Why was this change made? If the reason is to support decimals, I suggest using `DecimalPad` instead.

src/iOS/Xamarin.iOS/Samples/Map/ChangeAtmosphereEffect/ChangeAtmosphereEffect.cs
```diff
@@ -79,15 +79,13 @@ public override void LoadView()
             _atmosphereEffectPicker = new UISegmentedControl("Realistic", "Horizon only", "None");
             _atmosphereEffectPicker.TranslatesAutoresizingMaskIntoConstraints = false;
             _atmosphereEffectPicker.SelectedSegment = 1;
-            _atmosphereEffectPicker.TintColor = UIColor.White;
-            _atmosphereEffectPicker.BackgroundColor = UIColor.Black;
+            _atmosphereEffectPicker.BackgroundColor = UIColor.White;
```
nCastle1, line 82: This is a good change. I recommend updating the feature layer extrusion sample to have a white background as well.

src/iOS/Xamarin.iOS/Samples/Geometry/BufferList/BufferList.cs
ZackAllen: NumberPad doesnt have a return key. If I use numberpad or decimal pad I would have to add a done button to the UI for when the user is done typing.

PR #802, opened 2020-06-12
Local Server 100.8

ZackAllen:


mbranscomb:
@ZackAllen Does the same change apply to the WPF .NET core project?

ZackAllen:
> @ZackAllen Does the same change apply to the WPF .NET core project?

Yes it applies to the core project as well.

PR #819, opened 2020-08-26
Update 9 Fixes - Xamarin.Forms

ZackAllen:


src/Forms/Android/ArcGISRuntime.Xamarin.Forms.Android.csproj
```diff
@@ -118,7 +118,7 @@
       <Version>100.9.0</Version>
     </PackageReference>
     <PackageReference Include="MarkedNet">
-      <Version>2.1.2</Version>
```
nCastle1, line 121: I've forgotten to mention this in the past, but it is a good idea to update the SSG templates so that the nuget packages match what is being used by the samples.

PR #826, opened 2020-09-18
Update target version for UWP sample viewer

ZackAllen:


src/UWP/ArcGISRuntime.UWP.Viewer/ArcGISRuntime.UWP.Viewer.csproj
```diff
@@ -12,7 +12,7 @@
     <AssemblyName>ArcGISRuntime</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.17763.0</TargetPlatformVersion>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.18362.0</TargetPlatformVersion>
```
mbranscomb, line 15: I think this value depends if we consider the target version to be the recommended version or latest version tested. If that's the case, after 100.9 would it now be `19041` (Version 2004 / May 2020)?

nCastle1 (approved):

PR #830, opened 2020-12-07
Android scene dispose improvements

ZackAllen:


nCastle1 (approved):
Tested with HTC U11, scene memory management is much improved.

mstefarov:
The disposing code looks good. I'm not sure if GC code is really needed, if you are disposing manually. You could skip it.

ZackAllen:
@mstefarov I am open to changing this further before release.

PR #838, opened 2021-01-20
100.10 BasemapStyle update

ZackAllen:
Updated every sample to use the new basemap style pattern


src/Android/Xamarin.Android/Samples/Layers/DisplayKml/DisplayKml.cs
```diff
@@ -49,7 +49,7 @@ protected override void OnCreate(Bundle bundle)
         private void Initialize()
         {
             // Set up the basemap.
-            _mySceneView.Scene = new Scene(Basemap.CreateImageryWithLabels());
```
nCastle1, line 52: The `ArcGISImagery` layer is using a vector layer for labels, so that won't work in scenes.

I think ImageryStandard works without issues, but that is one of a very small handful that work. 

I'm not sure if we want to avoid use of `BasemapStyle` in scenes altogether, or just be careful to only use styles that are backed by raster layers.

src/Forms/Shared/Samples/Layers/OpenStreetMapLayer/OpenStreetMapLayer.xaml.cs
```diff
@@ -29,7 +29,7 @@ public OpenStreetMapLayer()
         private void Initialize()
         {
             // Create the OpenStreetMap basemap.
-            Basemap osmBasemap = Basemap.CreateOpenStreetMap();
```
nCastle1, line 32: Previously this sample was showing content from the OpenStreetMap service directly, now we're showing an ArcGIS service built with OSM data; we might want to think about what specifically this sample is meant to show. I suspect readme updates may be in order here as well.

nCastle1 (changes requested):
We'll need to figure out how to handle scenes - a few scene samples were using raster reference layers, and those aren't available through BasemapStyle enum.

src/Android/Xamarin.Android/Samples/Layers/DisplayKml/DisplayKml.cs
ZackAllen: These changes have been reverted. Plan is to move these to Basemap style when support for vector is added.

nCastle1 (approved):

PR #840, opened 2021-01-22
Update 10: Fixes for "Create and save map"

ZackAllen:
API key changes break the workflow of this sample. Removing the API key while using the sample is the preferred solution for now.


nCastle1 (approved):
Approved with a few suggestions, feel free to take or leave

PR #867, opened 2021-03-08
New sample: branch versioning

ZackAllen:


nCastle1 (changes requested):
I reviewed the code and tested the Android, WPF, and UWP implementations. let me know if you have any questions.

Things I found:

* Android, UWP, WPF screenshot, readme, metadata need to be updated
* Forms, iOS readme and metadata needs to be updated
* As written, the Android and WPF samples allow (and instruct) you to select and move a feature before creating the version, and I don't think that's how it is meant to be used. It probably makes sense to update the instruction label to clarify, and disable map interaction until after the version is created
* It seems the server remembers previously-created versions. Would it be possible to delete the version on the server before the sample runs? Otherwise it probably makes sense to use a random version name (e.g. a guid) to prevent conflicts.
* I got an error building the Forms UWP sample, have you seen that? Specifically I'm getting `The "WireUpCoreRuntime" task was not given a value for the required parameter "CoreRuntimeSDKLocation".` No worries if not, there might be something wrong with my device, and I'm due to re-image next week.

src/Forms/Shared/Samples/Data/EditBranchVersioning/EditBranchVersioning.xaml.cs
```diff
@@ -0,0 +1,361 @@
...
+                {
+                    ShowAlert("Please select an access level");
+                    return;
+                }
+
+                // Set the user defined name, access level and description as service version parameters
+                ServiceVersionParameters newVersionParameters = new ServiceVersionParameters();
+                newVersionParameters.Name = NameEntryBox.Text;
+                newVersionParameters.Access = (VersionAccess)AccessBox.SelectedItem;
+                newVersionParameters.Description = DescriptionBox.Text;
```
nCastle1, line 318: Interestingly, on Forms UWP, if the user leaves the description field blank, an ArgumentNullException is thrown here. Apparently the textbox text defaults to null rather than an empty string. Curiously, the argument null exception is thrown when the description is set, but not if the description is never set. I recommend checking for null here and using an empty string if it is `= DescriptionBox.Text ?? "";` or similar.

nCastle1 (changes requested):

nCastle1 (approved):

PR #872, opened 2021-04-09
Fixes for UWP toolkit 7.0.1

ZackAllen:
These fixes allow UWP to build and run with the latest toolkit package. I will create a separate PR with style fixes to the new tabs. ( will copy from WinUI )


nCastle1 (changes requested):
The UI suddenly looks very different for me, are you seeing this?

image

The background appearance for tabs is slightly awkward, specifically the boundary between the sample title and the tab area:

image

The tabs also have close buttons, which I think we don't want. They don't do anything. Same with the new tab button:

image

The 'About' and 'Third party software' tabs aren't working at all for me:

image

I got an NRE when coming back to the samples page from the about/settings page:

image

(only happens when coming back from settings after user was already on sample page)

Tabs are able to be dragged and re-arranged, which I don't think is intended. Dragging and re-arranging the sample page tabs is another trigger for NRE in the same place as above.

Source code preview doesn't appear to be working at all, and I don't see any output. We might have to revert to plaintext or update the relevant package.

nCastle1 (approved):

PR #878, opened 2021-04-22
Fix local server version

dotMorten:


mbranscomb (approved):
@dotMorten Thanks, approved.

PR #881, opened 2021-04-22
Local Server 100.10

ZackAllen:
-Updated the package reference and deployments file.
-Updated the data in the samples to use mpkx, gpkx, and tpkx.


dotMorten (approved):

mbranscomb (changes requested):
@ZackAllen Submitted a PR for your branch which updates the .NET Core project to v100.10 and corrects the deployment config file.

mbranscomb (approved):

PR #885, opened 2021-04-22
Remove redundant files

ZackAllen:


mbranscomb (approved):
@ZackAllen 
Files not required since pre v100.x.
Changes look good to me.

PR #891, opened 2021-05-18
Create & save map sample basemap updates

ZackAllen:


TADraeseke (commented):
Based on the review last week it looks like we might not be making this change after all and moving to just using a named user auth challenge. More to come...

PR #899, opened 2021-06-16
Label definition sample updates

ZackAllen:
* Updated `Display subtype feature layer` sample to use label definition constructed using API instead of JSON
* Updated `Show labels on layer` sample to use label definition constructed using Arcade expression instead of JSON


nCastle1 (approved):
Just one comment, otherwise looks good

PR #907, opened 2021-07-18
Add WinUI to metadata and sample creation scripts

ZackAllen:
This PR supports WinUI in the following ways:
- Adds support for generating WinUI samples with the `samplegen.py` script.
- Adds support for syncing WinUI metadata and json information with the `process_metadata.py` script.

This sample also:
- Updates sample metadata to use 100.11.2 instead of 100.11.0 across all platforms.


tools/sample_generator/templates/default/WinUI.xaml.cs
```diff
@@ -0,0 +1,44 @@
...
+    [offline_data_attr]
+    public partial class sample_name
+    {
+        public sample_name()
+        {
+            InitializeComponent();
+            Initialize();
+        }
+
+        private void Initialize()
```
mbranscomb, line 40: Should this be async Task?

tools/sample_generator/templates/default/WinUI.xaml
```diff
@@ -0,0 +1,10 @@
+<UserControl 
+    x:Class="ArcGISRuntime.WinUI.Samples.sample_name.sample_name"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:esri="using:Esri.ArcGISRuntime"
+    xmlns:esriUI="using:Esri.ArcGISRuntime.UI.Controls">
+    <Grid>
+        <esriUI:Geo_View x:Name="MyGeo_View" />
```
mbranscomb, line 8: Is `Geo_View` a placeholder?

mbranscomb (changes requested):
@ZackAllen Perhaps just a minor change to the default xaml.cs to change void to async Task.

tools/sample_generator/templates/default/WinUI.xaml
ZackAllen: Geo_View gets changed to MapView or SceneView when running the script.

tools/sample_generator/templates/default/WinUI.xaml.cs
ZackAllen: There aren't always async calls in initialize, so this should be changed when async calls are used.

mbranscomb: @ZackAllen That's a fair point. Seems like more often than not though, we're creating maps/scenes/layers/symbols and they're all async operations we should be awaiting (or handling load status changed).

nCastle1 (approved):

PR #909, opened 2021-08-05
New sample: Query with CQL filters

ZackAllen:


src/iOS/Xamarin.iOS/Samples/Layers/QueryCQLFilters/QueryCQLFilters.cs
```diff
@@ -0,0 +1,387 @@
...
+            _applyToolbar = new UIToolbar { TranslatesAutoresizingMaskIntoConstraints = false };
+            _applyToolbar.Items = new[]
+            {
+                _applyButton,
+                new UIBarButtonItem(UIBarButtonSystemItem.FlexibleSpace),
+                _resetButton
+            };
+
+            // Add the views.
+            View.AddSubviews(_myMapView, _activityIndicator, dateSwitchStack, maxFeaturesStack, clauseStack, datesStack, _applyToolbar);
```
nCastle1, line 303: Activity indicator should be added last to ensure that it is on top of the other views; currently it is possible to interact with the sample while things are loading.

src/Android/Xamarin.Android/Samples/Layers/QueryCQLFilters/QueryCQLFilters.cs
```diff
@@ -0,0 +1,354 @@
...
+            return null;
+        }
+
+        public QueryParameters CreateQueryParameters()
+        {
+            // Create new query parameters.
+            var queryParameters = new QueryParameters();
+
+            // Assign the where clause if one is provided.
+            if (_whereClauseSpinner.SelectedItem?.ToString() is string selectedClause) queryParameters.WhereClause = selectedClause;
```
nCastle1, line 172: This and the next `if` should be updated to use brackets

src/Android/Xamarin.Android/Samples/Layers/QueryCQLFilters/QueryCQLFilters.cs
```diff
@@ -0,0 +1,354 @@
...
+                _progressBar.Visibility = Android.Views.ViewStates.Gone;
+            }
+        }
+
+        private async void StartDateClick(object sender, EventArgs e)
+        {
+            try
+            {
+                DateTime promptDate = new DateTime(2011, 6, 13);
+                if (_startTime is DateTime validStart) promptDate = validStart;
```
nCastle1, line 273: I recommend using braces here. Also, I recommend adding comments to the next two methods.

src/Forms/Shared/Samples/Layers/QueryCQLFilters/QueryCQLFilters.xaml.cs
```diff
@@ -0,0 +1,217 @@
...
+                await Application.Current.MainPage.DisplayAlert("Error", ex.Message, "OK");
+            }
+            finally
+            {
+                // Update the UI.
+                LoadingProgressBar.IsVisible = false;
+            }
+        }
+
+        private Renderer GetRendererForTable(FeatureTable table)
```
nCastle1, line 117: Is this necessary? It seems like the result will always be a polyline.

nCastle1 (changes requested):
Overall, looks good, but I do have some UI & code feedback.

src/Forms/Shared/Samples/Layers/QueryCQLFilters/QueryCQLFilters.xaml.cs
ZackAllen: This was there in case the user swapped out default data for their own. Considering that they would have to change url, collection id, and where clauses anyway, I think its good to simplify this and let the user change the renderer if necessary.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Layers/QueryCQLFilters/QueryCQLFilters.xaml.cs
```diff
@@ -196,5 +177,13 @@ private async void ApplyQuery_Click(object sender, RoutedEventArgs e)
                 LoadingProgressBar.Visibility = Visibility.Hidden;
             }
         }
+
+        private void DateBox_Checked(object sender, RoutedEventArgs e)
```
nCastle1, line 181: This could be made a bit cleaner if done through binding in the XAML. Just a suggestion

nCastle1 (approved):

PR #912, opened 2021-08-12
EditFeatureAttachment tidy-up

mstefarov:
- Remove dependency to deprecated Xamarin.FilePicker plugin. This functionality is now built into Xamarin Essentials. See [migration guide](https://github.com/jfversluis/FilePicker-Plugin-for-Xamarin-and-Windows#migration-guide).
- Filter displayed attachments consistently, to only show jpeg/jpg. Previously the sample would sometimes show non-jpeg attachments after adding or removing an attachment. Now the filtering logic is centralized.


ZackAllen (approved):
I ran the auto-formatter scripts on the sample and xaml layout.

PR #916, opened 2021-08-13
XF: Use numeric keyboard for numeric entry fields

mstefarov:
This improves the sample experience of iOS and Android devices -- focusing these fields will now pop up the numeric keyboard. This is a best practice that gives app users a hint that only numbers are expected/allowed, and makes it quicker to enter numbers.


ZackAllen (approved):
I did some additional tweaking, some of these samples have very dated UI

PR #936, opened 2021-11-17
Updated WinUI to latest WinUI 1.0.0

dotMorten:
Also changed to simpler single-project


ZackAllen (approved):
Runs correctly on my machine.

PR #939, opened 2021-11-29
Winui toolkit update

dotMorten:
Updates to the newly released non-preview version of the community toolkit


ZackAllen (approved):
Builds and runs without issue.

PR #942, opened 2021-12-08
WinUI: Add remaining sample implementations

ZackAllen:


nCastle1 (changes requested):
- ApplyMosaicRule screenshot needs to be updated
- Possibly due to differences between UWP and WinUI, some of the sample UIs need to be updated. The spacing/alignment of BrowseOAFeatureService, EditBranchVersioning, and possibly others need review
- BrowseOAFeatureService.jpg has been added to the AddAnIntegratedMeshLayer sample folder. EditBranchVersioning.jpg has been added to the ApplyMosaicRule sample folder. There are other issues like that.

PR #949, opened 2021-12-21
Sample update: Display filters

ZackAllen:


src/UWP/ArcGISRuntime.UWP.Viewer/Samples/Layers/FeatureLayerDefinitionExpression/FeatureLayerDefinitionExpression.xaml.cs
```diff
@@ -1,79 +1,147 @@
...
+                {
+                    if (_timer == null)
+                    {
+                        _timer = new DispatcherTimer
+                        {
+                            Interval = TimeSpan.FromSeconds(1),
+                        };
+                        _timer.Tick += (a, b) =>
+                        {
+                            _ = QueryFeatureCountAsync();
```
nCastle1, line 114: I'd like to see a Cancellation-based workflow here, rather than the timer-based approach, or the stored task+bool to retry approach used on other platforms.

src/WinUI/ArcGISRuntime.WinUI.Viewer/Samples/Layers/DisplayOACollection/readme.metadata.json
```diff
@@ -17,7 +17,7 @@
         "web"
     ],
     "nuget_packages": {
-        "Esri.ArcGISRuntime": "100.12.0"
+        "Esri.ArcGISRuntime": "100.13.0"
```
nCastle1, line 20: It would be best to exclude these unrelated changes, and maybe open a second PR with metadata cleanup

nCastle1 (changes requested):

PR #951, opened 2021-12-27
Removed use of `Clone` in samples

ZackAllen:


nCastle1 (approved):

nixta:
I'm just looking at this PR and noticed that the Copyright years were updated. Is that correct? My understanding is that these should remain unchanged with the original year.

PR #960, opened 2022-01-27
New sample: Display device location with NMEA data sources

ZackAllen:


src/WinUI/ArcGISRuntime.WinUI.Viewer/MainPage.xaml.cs
```diff
@@ -81,12 +79,12 @@ private void LoadTreeView(SearchableTreeNode fullTree)
             }
             CategoriesTree.RootNodes.Clear();
 
-            muxc.TreeViewNode rootNode;
```
nCastle1, line 84: Was this intended for this PR?

nCastle1 (commented):
Sample looks good, I have some initial suggestions, but use your judgment.

src/WinUI/ArcGISRuntime.WinUI.Viewer/MainPage.xaml.cs
ZackAllen: This is some minor cleanup I did while editing this file. That import and prefix are not needed in WinUI, I believe they were in UWP.

src/Forms/Shared/Samples/Location/LocationWithNMEA/LocationWithNMEA.xaml.cs
```diff
@@ -0,0 +1,211 @@
...
+            _nmeaSource = new NmeaLocationDataSource(SpatialReferences.Wgs84);
+
+            // Android and UWP:
+            // To create a NmeaLocationDataSource for a bluetooth device, use the `FromBluetooth` constructor. https://developers.arcgis.com/net/api-reference/api/android/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Location.NmeaLocationDataSource.FromBluetooth.html
+            // To create a NmeaLocationDataSource from a serial port, use the `FromSerialPort` constructor. https://developers.arcgis.com/net/api-reference/api/android/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Location.NmeaLocationDataSource.FromSerialPort.html
+
+            // iOS:
+            // When using an NMEA device on iOS, use the `NmeaLocationDataSource.FromAccessory` constructor. https://developers.arcgis.com/net/api-reference/api/ios/Esri.ArcGISRuntime/Esri.ArcGISRuntime.Location.NmeaLocationDataSource.FromAccessory.html
+
+            // Set the location data source to use the stream from our simulator.
```
nCastle1, line 62:
```suggestion
            // Set the location data source to use the stream from the simulator.
```

src/Forms/Shared/Samples/Location/LocationWithNMEA/readme.md
```diff
@@ -0,0 +1,61 @@
...
+
+* com.amanenterprises.nmeasource
+* com.dualav.xgps150
+* com.garmin.pvt
+* com.junipersys.geode
+* com.leica-geosystems.zeno.gnss
+* com.searanllc.serial
+* com.trimble.correction, com.trimble.command (1)
+
+(1) Some Trimble models requires a proprietary SDK for NMEA output.
```
nCastle1, line 57:
```suggestion
(1) Some Trimble models require a proprietary SDK for NMEA output.
```

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Location/LocationWithNMEA/LocationWithNMEA.xaml.cs
```diff
@@ -0,0 +1,209 @@
...
+            _lineCounter = 0;
+        }
+
+        private void TimerElapsed(object sender, ElapsedEventArgs e)
+        {
+            // Get the next NMEA string and append return and newline characters to it.
+            string nmeaString = $"{_nmeaStrings[_lineCounter]}\r\n";
+
+            // Check if the start of a new location message.
+            if (nmeaString.StartsWith("GPGGA"))
```
nCastle1, line 187: This isn't working, the messages start with `$`

nCastle1 (changes requested):
I made a few suggestions, those suggestions apply to other platforms.

nCastle1 (changes requested):
I'd like to see two improvements to make this easier to use:

- Speed up the interval or update the data so that changes happen more frequently. It looked like it was broken at first, but it was just taking a long time to update. Maybe, you could show the latest NMEA message. Maybe a speed increase would be enough.
- Add an improved help label indicating the status: not started, waiting for first message, etc.

Optional suggestions:

- Maybe Recenter should just be on by default? I don't know if the extra button is necessary, especially since the demo doesn't work very well without recenter on.
- Reset should clear all of the labels

nCastle1 (commented):
I'm seeing strange behavior on Forms UWP. UI tends to resize a lot due to the length of the label changing. That can be fixed by setting `LineBreakMode` to `TailTruncation`. 

Rather than disabling panning, I recommend re-enabling autopan mode when the location updates. I thought something was broken at first.

nCastle1 (approved):

PR #963, opened 2022-02-08
New Sample: Browse Building Floors

asantana917:
This is the new branch after leaving the old branch.


src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml
```diff
@@ -0,0 +1,19 @@
+﻿<UserControl x:Class="ArcGISRuntime.WPF.Samples.BrowseBuildingFloors.BrowseBuildingFloors"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:esri="http://schemas.esri.com/arcgis/runtime/2013">
+    <Grid>
+        <esri:MapView x:Name="MyMapView"/>
+        <Border Style="{StaticResource BorderStyle}" Margin="0,30,30,0" Width="144">
```
nCastle1, line 7: BorderStyle already defines a margin, so that can be left out.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                // Get the portal item for a web map using its unique item id.
+                PortalItem mapItem = await PortalItem.CreateAsync(portal, _portalItem);
+                Map map = new Map(mapItem);
+
+                MyMapView.Map = map;
+
+                // Map needs to be loaded in order for floormanager to be used.
+                await MyMapView.Map.LoadAsync();
+                await MyMapView.Map.FloorManager.LoadAsync();
```
nCastle1, line 56: FloorManager happens to not be null for this web map, but it could be null for others.

I'd prefer an explicit check to make sure the floor manager isn't null, rather than implicitly relying on an exception being thrown.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                if (MyMapView.Map.FloorManager.LoadStatus == LoadStatus.Loaded)
+                {
+                    _floorManager = MyMapView.Map.FloorManager;
+
+                    // Use the dictionary to add the level's name as the key and the FloorLevel object with the associated level's name.
+                    foreach (FloorLevel level in _floorManager.Levels)
```
nCastle1, line 63: This code is assuming that there is only one facility, so it is appropriate to add all levels to the floor picker. In many real-world use cases, floor-aware maps will have multiple facilities, and there may even be multiple sites.

The shown approach wouldn't work well if you were to connect this to the Esri campus web map - there would be  dozens of floors. Typically the expectation for a floor picker is that there will only be one facility/building in the floor picker at a time.

To fix this, I recommend adding code here to explicitly choose the first facility, and use the facility's `Levels` collection, rather than the root FloorManager levels collection.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                }
+                // Provides an error message if the floor manager failed to load.
+                else if (MyMapView.Map.FloorManager.LoadStatus == LoadStatus.FailedToLoad)
+                {
+                    MessageBox.Show("Floor manager failed to load.", "Error", MessageBoxButton.OK, MessageBoxImage.Error);
+                    return;
+                }
+
+                FloorChooser.ItemsSource = _floorOptions.Keys;
```
nCastle1, line 75: With the change mentioned in the XAML file, the loop above can be removed, as can the `_floorOptions` dictionary. Instead, you can do `FloorChooser.ItemsSource = _selectedFacility.Levels` and use the `DisplayMemberPath` or a `DataTemplate` as needed to customize how it is displayed.

src/Forms/Shared/Forms.projitems
```diff
@@ -121,8 +122,8 @@
     <None Include="$(MSBuildThisFileDirectory)Samples\Geoprocessing\AnalyzeHotspots\AnalyzeHotspots.jpg" />
     <None Include="$(MSBuildThisFileDirectory)Samples\Geoprocessing\AnalyzeViewshed\AnalyzeViewshed.jpg" />
     <None Include="$(MSBuildThisFileDirectory)Samples\Geoprocessing\ListGeodatabaseVersions\ListGeodatabaseVersions.jpg" />
-    <None Include="$(MSBuildThisFileDirectory)Samples\GraphicsOverlay\AddGraphicsRenderer\AddGraphicsRenderer.jpg" />
```
ZackAllen, line 124: This change should be undone.

src/Forms/Shared/Samples/Map/BrowseBuildingFloors/readme.md
```diff
@@ -0,0 +1,43 @@
...
+
+## How it works
+
+1. Create a `PortalItem` using the identifier of a floor-aware web map.
+2. Create a map using the portal item.
+3. Create a map view and assign the map to it.
+4. Wait for the map to load and retrieve the map's `FloorManager` property.
+5. Wait for the floor manager to load and retrieve the floor-aware data. 
+6. Set all floors to not visible.
+7. Set only the selected `FloorLevel` to visible using the `isVisible` property of the floor level.
```
ZackAllen, line 23:
```suggestion
7. Set only the selected `FloorLevel` to visible using the `IsVisible` property of the floor level.
```

Property is capitalized in .NET.

src/Forms/Shared/Forms.projitems
```diff
@@ -1607,4 +1615,9 @@
     <EmbeddedResource Include="$(MSBuildThisFileDirectory)Resources\SettingsPage\key.png" />
     <EmbeddedResource Include="$(MSBuildThisFileDirectory)Resources\SettingsPage\keygrey.png" />
   </ItemGroup>
+  <ItemGroup>
```
ZackAllen, line 1618: Any diff involving AddGraphicsRenderer should be undone.

src/WPF/ArcGISRuntime.WPF.Viewer.Net.sln
```diff
@@ -39,8 +39,14 @@ Global
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|Any CPU.Deploy.0 = Debug|Any CPU
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|x64.ActiveCfg = Debug|x64
+		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|x64.Build.0 = Debug|x64
```
ZackAllen, line 42: Changes to the .sln should not be included.

src/WinUI/ArcGISRuntime.WinUI.Viewer/ArcGISRuntime.WinUI.Viewer.csproj
```diff
@@ -54,11 +54,12 @@
     <PackageReference Include="Esri.ArcGISRuntime.Hydrography" Version="100.13.0" />
```
ZackAllen, line 54: These nuget changes should not be in this PR.

ZackAllen (changes requested):

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+
+        private void OnFloorChooserSelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            // Get the name of the selected floor.
+            string selectedFloorName = e.AddedItems[0].ToString();
+
+            // Set all existing floors visibility to false.
+            foreach (FloorLevel level in _floorManager.Levels)
```
nCastle1, line 90: It might make sense to be very clear about the design choice here. This code will hide all levels in all facilities in all sites before showing the selected level. In a lot of scenarios, it would make sense to keep all of the facilities in sync, so that if you have two buildings side-by-side, when you select level 2 in one, you see level 2 in all.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            InitializeComponent();
+            _ = Initialize();
+        }
+
+        private async Task Initialize()
+        {
+            try
+            {
+                ArcGISPortal portal = await ArcGISPortal.CreateAsync();
```
nCastle1, line 46: I know this is specified in the design, but there might be too much emphasis on the portal here. Floor-aware maps (and scenes) are really about web maps and web scenes, so IMO it would make sense to use the map constructor taking a URL here.

nCastle1 (changes requested):
This is a great start. See comments for specific suggestions

src/Forms/Shared/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+            // Set all existing floors visibility to false.
+            foreach (FloorLevel level in _floorManager.Levels)
+            {
+                level.IsVisible = false;
+            }
+
+            // Set the selected floor visibility to true.
+            _floorOptions[selectedFloorName].IsVisible = true;
+
+            MyMapView.SetViewpoint(MyMapView.Map.InitialViewpoint);
```
nCastle1, line 98: Forgot to mention, both levels and facilities can be expected to have geometries that can be used for setting the viewpoint. It would be most ideal if we could demonstrate that through this sample.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
asantana917: @nCastle1 I have the code also checking if the the FloorManager fails to load. Should I add an extra term to check if the FloorManger is null as well?

nCastle1: Yes, this line will throw if you try with a map that isn't floor-aware because FloorManager will still be null

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+        name: "Browse building floors",
+        category: "Map",
+        description: "Display and browse through building floors from a floor-aware web map.",
+        instructions: "Use the combo box to browse different floor levels in the facility. Only the selected floor will be displayed.",
+        tags: new[] { "building", "facility", "floor", "floor-aware", "floors", "ground floor", "indoor", "level", "site", "story" })]
+    public partial class BrowseBuildingFloors
+    {
+        private const string _floorData = @"https://ess.maps.arcgis.com/home/item.html?id=f133a698536f44c8884ad81f80b6cfc7";
+        private FloorManager _floorManager;
+        private FloorFacility _selectedFacilities;
```
nCastle1, line 32: Since there's only one, this should be renamed to `_selectedFacility`.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    [ArcGISRuntime.Samples.Shared.Attributes.Sample(
+        name: "Browse building floors",
+        category: "Map",
+        description: "Display and browse through building floors from a floor-aware web map.",
+        instructions: "Use the combo box to browse different floor levels in the facility. Only the selected floor will be displayed.",
+        tags: new[] { "building", "facility", "floor", "floor-aware", "floors", "ground floor", "indoor", "level", "site", "story" })]
+    public partial class BrowseBuildingFloors
+    {
+        private const string _floorData = @"https://ess.maps.arcgis.com/home/item.html?id=f133a698536f44c8884ad81f80b6cfc7";
```
nCastle1, line 30: I don't think the `@` is necessary here.

Also, since the item is public, I don't think there's a need to highlight the ESS org specifically; URL can just be `https://www.arcgis.com/home/item.html?id=f133a698536f44c8884ad81f80b6cfc7`

src/WPF/ArcGISRuntime.WPF.Viewer.Net.sln
```diff
@@ -39,8 +39,13 @@ Global
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|Any CPU.Deploy.0 = Debug|Any CPU
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|x64.ActiveCfg = Debug|x64
+		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|x64.Deploy.0 = Debug|x64
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|x86.ActiveCfg = Debug|x86
+		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|x86.Build.0 = Debug|x86
+		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Debug|x86.Deploy.0 = Debug|x86
 		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{13A60801-7B9A-4573-923F-9F2373E8BF84}.Release|Any CPU.Build.0 = Release|Any CPU
```
nCastle1, line 47: These changes and the WinUI package changes should be reverted.

src/WinUI/ArcGISRuntime.WinUI.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+        description: "Display and browse through building floors from a floor-aware web map.",
+        instructions: "Use the combo box to browse different floor levels in the facility. Only the selected floor will be displayed.",
+        tags: new[] { "building", "facility", "floor", "floor-aware", "floors", "ground floor", "indoor", "level", "site", "story" })]
+    public partial class BrowseBuildingFloors
+    {
+        private const string _floorData = @"https://ess.maps.arcgis.com/home/item.html?id=f133a698536f44c8884ad81f80b6cfc7";
+        private FloorManager _floorManager;
+        private FloorFacility _selectedFacilities;
+
+        // Collection of floors.
```
nCastle1, line 33: This can be removed

nCastle1 (changes requested):

src/Forms/Shared/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,108 @@
...
+                if (MyMapView.Map.FloorDefinition != null)
+                {
+                    await Application.Current.MainPage.DisplayAlert("Alert", "The layer is not floor aware.", "OK");
+                    return;
+                }
+
+                await MyMapView.Map.FloorManager.LoadAsync();
+
+                // Checks to see if the floormanager loaded and check to see if there is a floormanager.
+                if (MyMapView.Map.FloorManager.LoadStatus == LoadStatus.Loaded && MyMapView.Map.FloorManager != null)
```
nCastle1, line 64: Since the `LoadAsync` call is awaited, I think you can count on any load errors to throw an exception, which means you don't need to check the load status here or in the `else if`. Also, with the order the two conditions are written, if the FloorManager is null here, you would get a `NullReferenceException` before the null check could happen.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Map/BrowseBuildingFloors/BrowseBuildingFloors.xaml.cs
```diff
@@ -0,0 +1,94 @@
...
+{
+    [ArcGISRuntime.Samples.Shared.Attributes.Sample(
+        name: "Browse building floors",
+        category: "Map",
+        description: "Display and browse through building floors from a floor-aware web map.",
+        instructions: "Use the combo box to browse different floor levels in the facility. Only the selected floor will be displayed.",
+        tags: new[] { "building", "facility", "floor", "floor-aware", "floors", "ground floor", "indoor", "level", "site", "story" })]
+    public partial class BrowseBuildingFloors
+    {
+        private const string _floorData = "https://ess.maps.arcgis.com/home/item.html?id=f133a698536f44c8884ad81f80b6cfc7";
```
nCastle1, line 30: I think this might be confusing, since the constant is holding a string with a URL, not floor data.

Since the URL isn't referenced anywhere else, the simplest solution is probably to put the string literal in the URL constructor on line 43.

nCastle1 (changes requested):

nCastle1 (commented):
This is looking pretty good. My only remaining suggestion is that the WinUI UI should be tweaked (maybe just center or stretch the combobox?) and the screenshot should be updated; it looks like the WPF screenshot as copied.

ZackAllen (approved):

PR #978, opened 2022-03-02
WPF store package build fixes

ZackAllen:
- The use of `delegate` in the new NMEA sample breaks the store package build.
- The store package had an issue with an x86 warning that needs to be ignored.

### DIFF
--- a/src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Location/LocationWithNMEA/LocationWithNMEA.xaml.cs
+++ b/src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Location/LocationWithNMEA/LocationWithNMEA.xaml.cs
@@ -81,10 +81,10 @@ namespace ArcGISRuntime.WPF.Samples.LocationWithNMEA
 
         private void UpdateNmeaMessageLabel(object sender, NmeaMessageEventArgs e)
         {
-            Dispatcher.BeginInvoke(delegate ()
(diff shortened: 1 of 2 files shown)


src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Location/LocationWithNMEA/LocationWithNMEA.xaml.cs
```diff
@@ -81,10 +81,10 @@ private async Task Initialize()
 
         private void UpdateNmeaMessageLabel(object sender, NmeaMessageEventArgs e)
         {
-            Dispatcher.BeginInvoke(delegate ()
+            Dispatcher.BeginInvoke(new Action(() =>
```
nCastle1, line 84: The `new Action(` isn't needed here; you can provide the anonymous method directly.

nCastle1 (approved):
Please try the suggested change and make that change if it works with the store package project.

src/WPF/ArcGISRuntime.WPF.Viewer/Samples/Location/LocationWithNMEA/LocationWithNMEA.xaml.cs
ZackAllen: I get the same build failure with the `() =>` pattern. Keeping this as action.

PR #981, opened 2022-03-07
Enhance "Custom dictionary style" to also use new DictionarySymbolStyle constructor from a portal item

duffh:


src/WinUI/ArcGISRuntime.WinUI.Viewer/Samples/Symbology/CustomDictionaryStyle/readme.md
```diff
@@ -10,19 +10,22 @@ When symbolizing geoelements in your map, you may need to convey several pieces
...
+2. Use `DictionarySymbolStyle.CreateFromFileAsync()` to create a `DictionarySymbolStyle` from a .stylx file.
+3. Create a `DictionarySymbolStyle` with a `PortalItem`.
+4. Create and apply the desired `DictionaryRenderer` to the `FeatureLayer` for each `DictionarySymbolStyle` by selecting the appropriate radio button.
 
 ## Relevant API
 
 * DictionaryRenderer
 * DictionarySymbolStyle
 * DictionarySymbolStyleConfiguration
+* Portal
```
nCastle1, line 27: I don't think `Portal` is necessary here. The PortalItem is used but the Portal is only used as an input to the item. I think the PortalItem could have been created without a Portal object.

src/Forms/Shared/Samples/Symbology/CustomDictionaryStyle/CustomDictionaryStyle.xaml.cs
```diff
@@ -30,45 +32,83 @@ public partial class CustomDictionaryStyle : ContentPage
...
+                PortalItem portalItem = await PortalItem.CreateAsync(portal, "adee951477014ec68d7cf0ea0579c800");
 
-                // Set the restaurant layer renderer to the dictionary renderer.
-                restaurantLayer.Renderer = dictRenderer;
+                // Open the portal custom style file.
+                _restaurantStyleFromPortal = await DictionarySymbolStyle.OpenAsync(portalItem);
 
                 // Set the map's initial extent to that of the restaurants.
-                map.InitialViewpoint = new Viewpoint(restaurantLayer.FullExtent);
+                map.InitialViewpoint = new Viewpoint(_restaurantLayer.FullExtent);
```
nCastle1, line 73: I wonder if it would be better to re-order this code to group the map manipulation at the start, then set up the styles and renderers after.

src/Forms/Shared/Samples/Symbology/CustomDictionaryStyle/CustomDictionaryStyle.xaml.cs
```diff
@@ -30,45 +32,83 @@ public partial class CustomDictionaryStyle : ContentPage
         // Uri for the restaurants feature service.
         private readonly Uri _restaurantUri = new Uri("https://services2.arcgis.com/ZQgQTuoyBrtmoGdP/arcgis/rest/services/Redlands_Restaurants/FeatureServer/0");
 
+        // Hold a reference to the custom dictionary styles for use in the event handlers. 
+        private DictionarySymbolStyle _localRestaurantStyle;
+        private DictionarySymbolStyle _restaurantStyleFromPortal;
```
nCastle1, line 37: I wonder if it would set a better performance example to create and store the renderer, rather than re-creating it from the style each time?

src/Forms/Shared/Samples/Symbology/CustomDictionaryStyle/CustomDictionaryStyle.xaml.cs
```diff
@@ -30,45 +32,83 @@ public partial class CustomDictionaryStyle : ContentPage
...
             {
-                Console.WriteLine(ex.Message);
+                await Application.Current.MainPage.DisplayAlert("Error", ex.Message, "OK");
             }
         }
+
+        private void LocalStyleButtonClicked(object sender, EventArgs e)
+        {
+            // Create the dictionary renderer with the style file and the field overrides.
+            DictionaryRenderer dictionaryRenderer = new DictionaryRenderer(_localRestaurantStyle);
```
nCastle1, line 93: To make the sample code easier to comprehend, I think it would be better to have separate `LoadLocalStyle` and `LoadWebStyle` methods. Those two steps are be the core of the sample that users are interested in, so it will be good to isolate that from the button and map setup overhead.

src/Forms/Shared/Samples/Symbology/CustomDictionaryStyle/CustomDictionaryStyle.xaml
```diff
@@ -1,7 +1,37 @@
...
+        </Grid.ColumnDefinitions>
+        <esriUI:MapView x:Name="MyMapView"
+                        Grid.Row="0"
+                        Grid.Column="0"
+                        Grid.ColumnSpan="3" />
+        <Label Grid.Row="1"
+               Grid.Column="1"
+               Padding="0,0,0,5"
+               Text="Select Dictionary Symbol Style Source" />
+        <Button x:Name="LocalStyleButton"
```
nCastle1, line 25: Double check that radio button isn't available for Xamarin.Forms, since that might simplify some of the code behind.

nCastle1 (changes requested):
Sample enhancement looks great. I'd like to see some minor restructuring now that the complexity has been increased, to better highlight the parts users will be most interested in.

I also noticed a platform difference with how Xamarin.Forms is using buttons. I see that `RadioButton` exists for Xamarin.Forms, so it would be best to be consistent unless there's a reason not to.

asantana917 (commented):
Maybe this is a little nit picky for the screenshots but zoom in a little more so the icons aren't too meshed together.

asantana917 (approved):

ZackAllen (approved):

PR #984, opened 2022-03-15
Update Nearest Vertex sample to give correct planar distance result.

duffh:


src/Forms/Shared/Samples/Geometry/NearestVertex/NearestVertex.xaml.cs
```diff
@@ -41,38 +42,48 @@ public NearestVertex()
...
+            // Create the graphics overlay and set the selection color.
             _graphicsOverlay = new GraphicsOverlay();
 
-            // Add the overlay to the MapView
+            // Add the overlay to the MapView.
             MyMapView.GraphicsOverlays.Add(_graphicsOverlay);
 
-            // Create the point collection that defines the polygon
-            PointCollection polygonPoints = new PointCollection(SpatialReferences.WebMercator)
+            // Create the point collection that defines the polygon.
```
nCastle1, line 64: This is an unusual spatial reference choice for the sample. My understanding is that this was deliberate, so that when the planar distance calculation is done later, the results will accurately reflect the real-world distance? That should be mentioned in the comment here.

src/Forms/Shared/Samples/Geometry/NearestVertex/readme.metadata.json
```diff
@@ -26,7 +26,7 @@
         "/net/latest/forms/sample-code/nearestvertex.htm"
     ],
     "relevant_apis": [
-        "GeometryEngine",
+        "Geometry",
```
nCastle1, line 29: Metadata change should be reverted - sample is still using the `GeometryEngine` API.

src/Forms/Shared/Samples/Geometry/NearestVertex/readme.md
```diff
@@ -10,21 +10,28 @@ Determine the shortest distance between a location and the boundary of an area.
...
 2. Call `GeometryEngine.NearestVertex(inputGeometry, point)`.
-3. Use the returned `ProximityResult` to get the `Point` representing the polygon vertex, and to determine the distance between that vertex and the clicked point.
+3. Use the returned `ProximityResult` to get the `Point` representing the polygon vertex, and to determine the distance between that vertex and the tapped point.
 4. Call `GeometryEngine.NearestCoordinate(inputGeometry, point)`.
-5. Use the returned `ProximityResult` to get the `Point` representing the coordinate on the polygon, and to determine the distance between that coordinate and the clicked point.
+5. Use the returned `ProximityResult` to get the `Point` representing the coordinate on the polygon, and to determine the distance between that coordinate and the tapped point.
 
 ## Relevant API
 
+* Geometry
```
nCastle1, line 25: I don't think the code changes make this change necessary.

src/Forms/Shared/Samples/Geometry/NearestVertex/readme.md
```diff
@@ -10,21 +10,28 @@ Determine the shortest distance between a location and the boundary of an area.
...
 * GeometryEngine
 * ProximityResult
 
+## Additional information
+
+The value of `ProximityResult.distance` is planar (Euclidean) distance. Planar distances are only accurate for geometries that have a defined projected coordinate system, which preserves distance and length. The geometries in this sample are defined in Auxiliary Sphere (WKID 3857), which doesn't preserve distance. Hence we need to use the geodetic distance method to measure accurate distances in this example.
+
+Because the Web Mercator projected coordinate system is so common in web maps and it heavily distorts both area and length measurements, all geodetic measurement APIs are designed to allow geometries with Web Mercator spatial references as input. This allows you to measure geometries accurately in Web Mercator without having to project them to a different coordinate system.
```
nCastle1, line 33: This is a bit confusing since the sample doesn't actually show any geodetic APIs. I like Java's code better for this reason:

> The value of `ProximityResult.distance` is planar (Euclidean) distance. Planar distances are only accurate for geometries that have a defined projected coordinate system, which maintain the desired level of accuracy. The example polygon in this sample is defined in California State Plane Coordinate System - Zone 5 (WKID 2229), which maintains accuracy near Southern California. Accuracy declines outside the state plane zone.

nCastle1 (changes requested):
I understand that Runtime is planning on adding a geodetic version of the nearest coordinate method in a future release. I think it would have been better to make this change for v.next and update the sample design to take advantage of that new API.

PR #1004, opened 2022-04-07
Fixed bug where some category paths would be invalid

ZackAllen:
The old logic would change "GraphicsOverlay" to "Graphicsoverlay" with the title case call. This broke paths to readmes on Forms.iOS and potentially had other issues.


duffh:
This change looks good to me but I'm not 100% sure it fixes the problem you have identified, could you explain this solution?

duffh (approved):

asantana917 (approved):

PR #1023, opened 2022-05-09
New sample: Show device location using indoor positioning

ZackAllen:
# Description

Summary of the change and any relevant info.

## Type of change

- [x] New sample implementation

**Platforms tested on**:

- [x] Xamarin.Forms Android
- [ ] Xamarin.Forms iOS
- [x] Xamarin.Forms UWP

## Checklist

- [x] Runs and compiles on all active platforms
- [x] Branch is up to date with the latest main/v.next
- [x] All merge conflicts have been resolved
- [x] Self-review of changes
- [x] `sample_sync.py` runs without making changes
- [x] Code is commented with correct formatting
- [x] All variable and method names are good and make sense
- [x] There is no leftover commented code
- [x] Screenshots are correct size and display in description tab


src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+        {
+            // Get the properties from the new location.
+            IReadOnlyDictionary<string, object> locationProperties = loc.AdditionalSourceProperties;
+
+            locationProperties.TryGetValue(LocationSourcePropertyKeys.Floor, out object floor);
+            locationProperties.TryGetValue(LocationSourcePropertyKeys.PositionSource, out object positionSource);
+            locationProperties.TryGetValue(LocationSourcePropertyKeys.SatelliteCount, out object satCount);
+            locationProperties.TryGetValue("transmitterCount", out object transmitterCount);
+
+            int newFloor = int.Parse(floor.ToString());
```
dotMorten, line 163: `floor` can be null here, so this would throw null reference exception.

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+            // Set text for satellites or beacons.
+            string countText = string.Empty;
+            if (positionSource.Equals("GNSS"))
+            {
+                countText = $"Satellite count: {satCount}";
+            }
+            else if (positionSource.Equals("BLE") || positionSource.Equals("AppleIPS"))
+            {
+                countText = $"Beacon count: {transmitterCount}";
+            }
```
dotMorten, line 189: Rather than relying on the value of position source, you could just rely on the presence of "satCount" and "transmitterCount"? It would also be more future proof, and be able to display a combined set of values (a location could be based on both satellites and beacons)
Ie
```cs
if(floor != null) countText += $"Floor : {floor}\n";
if(satCount != null) countText += $"Satellite count: {satCount}\n";
if(transmitterCount != null) countText += $"Beacon count: {transmitterCount}\n";
```

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+                        return;
+                    }
+                }
+                catch
+                {
+                    return;
+                }
+
+#if __ANDROID__
+                // Get bluetooth permission for Android devices. Devies running Android 12 or higher need the `BluetoothScan` permission. Android 11 and below require the `Bluetooth` and `BluetoothAdmin` permissions.
```
dotMorten, line 98: Typo: `devies`

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+                try
+                {
+                    // WARNING: Never hardcode login information in a production application. This is done solely for the sake of the sample.
+                    return await AuthenticationManager.Current.GenerateCredentialAsync(info.ServiceUri, sampleUser, samplePass, info.GenerateTokenOptions);
+                }
+                catch (Exception ex)
+                {
+                    Debug.WriteLine(ex.Message);
+                    return null;
+                }
```
dotMorten, line 78: Since this service is for sample purposes, could we ask for an unsecured one, just to simplify the sample?

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+                QueryParameters queryParameters = new QueryParameters
+                {
+                    MaxFeatures = 1,
+                    // "1=1" will give all the features from the table.
+                    WhereClause = "1=1",
+                };
+                queryParameters.OrderByFields.Add(new OrderBy(dateCreatedFieldName.Name, SortOrder.Descending));
+
+                FeatureQueryResult queryResult = await positioningTable.QueryFeaturesAsync(queryParameters);
+                Guid globalID = (Guid)queryResult.First().Attributes["GlobalID"];
```
dotMorten, line 136: It's not clear to me why the sample is doing this. Could we explain why we're just making a query and grabbing whatever guid comes back first? Will the datasource not work without this guid?

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+                MyMapView.Map = new Map(item);
+                await MyMapView.Map.LoadAsync();
+
+                // Get the positioning table from the map.
+                await Task.WhenAll(MyMapView.Map.Tables.Select(table => table.LoadAsync()));
+                FeatureTable positioningTable = MyMapView.Map.Tables.FirstOrDefault(table => table.TableName.Equals(PositioningTableName));
+                if (positioningTable == null) return;
+
+                // Get a table of all of the indoor pathways.
+                FeatureLayer pathwaysFeatureLayer = MyMapView.Map.OperationalLayers.FirstOrDefault(l => l.Name.Equals(PathwaysLayerName, StringComparison.InvariantCultureIgnoreCase)) as FeatureLayer;
```
dotMorten, line 121: Use OfType to avoid cast and limit lookup `MyMapView.Map.OperationalLayers.OfType<FeatureLayer>()`

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+                ArcGISFeatureTable pathwaysTable = pathwaysFeatureLayer.FeatureTable as ArcGISFeatureTable;
```
dotMorten, line 122: `pathwaysFeatureLayer` could be null here, but there's no null-check. If you don't want to risk getting a null, use `Single()` instead of `FirstOrDefault()` in the line above (it'll then throw if it doesn't find the layer, and only one layer)

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+            }
+            else if (positionSource.Equals("BLE") || positionSource.Equals("AppleIPS"))
+            {
+                countText = $"Beacon count: {transmitterCount}";
+            }
+
+            // Update UI on the main thread.
+            Device.BeginInvokeOnMainThread(() =>
+            {
+                PositioningLabel.Text = $"Floor: {floor}\nPosition-source: {positionSource}\nHorizontal-accuracy: {string.Format("{0:0.##}", loc.HorizontalAccuracy)}m\n{countText}";
```
dotMorten, line 194: The label shows "IPS not started" all the way until the first location has been received (which might be never). This threw me off for a while. Could you also update the status to "Loading map", "Starting IPS", and "Waiting for location" in the initialize code?

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+                {
+                    return;
+                }
+
+#if __ANDROID__
+                // Get bluetooth permission for Android devices. Devies running Android 12 or higher need the `BluetoothScan` permission. Android 11 and below require the `Bluetooth` and `BluetoothAdmin` permissions.
+                bool bluetoothScanGranted = await MainActivity.Instance.AskForBluetoothPermission();
+                if (!bluetoothScanGranted)
+                {
+                    await Application.Current.MainPage.DisplayAlert("Error", "Bluetooth permission is required for use of indoor positioning.", "OK");
```
dotMorten, line 102: You could just throw, and the display alert code further down will handle it

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+                    if (status != Xamarin.Essentials.PermissionStatus.Granted)
+                    {
+                        await Application.Current.MainPage.DisplayAlert("Error", "Location permission required for use of indoor positioning.", "OK");
+                        return;
+                    }
+                }
+                catch
+                {
+                    return;
+                }
```
dotMorten, line 95: Instead of try/catch and displaying alert, you could just throw, and the display alert code further down will handle it - this would reduce code duplication.

src/Forms/Shared/Samples/Location/IndoorPositioning/IndoorPositioning.xaml.cs
```diff
@@ -0,0 +1,204 @@
...
+
+#if __ANDROID__
```
dotMorten, line 97: Is there a way to do this in a .NET Standard way? Most XF apps will be built against netstandard, and `__ANDROID__` wouldn't do anything.

ZackAllen: `if (Device.RuntimePlatform == Device.Android)` would work, but that doesn't solve us calling main activity. We use platform conditionals a lot in the Forms samples, eliminating them would be bigger than the scope of this sample.

duffh (approved):

nCastle1 (approved):

dotMorten (approved):
👍 🔥

PR #1127, opened 2022-10-31
MAUI: Updated scripts used in sample creation

duffh:
# Description

Updated scripts to include MAUI samples in sample generation and maintenance. 

## Type of change

- [ ] Bug fix
- [ ] New sample implementation
- [ ] Sample viewer enhancement
- [x] Other enhancement


tools/metadata_tools/process_metadata.py
```diff
@@ -256,9 +200,6 @@ def main():
                     list_of_samples[sample.category].append(sample)
                 else:
                     list_of_samples[sample.category] = [sample]
-        # write out samples TOC
-        if platform != "FormsAR":
-            write_samples_toc(get_platform_samples_root(platform, sample_root), get_relative_path_to_samples_from_platform_root(platform), list_of_samples)
```
ZackAllen, line 257: Why was this removed?

tools/metadata_tools/process_metadata.py
```diff
@@ -235,8 +181,6 @@ def main():
                     print(f"skipping path; does not exist: {path_to_readme}")
                     continue
                 sample.populate_from_readme(platform, path_to_readme)
-                if platform == "FormsAR":
```
ZackAllen, line 234: We shouldn't do this same thing with the maui AR samples. They should be put in normal category and we can just have platform specific exclusions for MacCatalyst and WinUI.

tools/metadata_tools/process_metadata.py
```diff
@@ -175,7 +175,7 @@ def main():
             d.sort()
             for sample_dir in d:
 
-                if 'resources' in sample_dir:
+                if sample_dir == 'resources':
```
ZackAllen, line 178: This could use a comment explaining the case with Sketch on map.

ZackAllen (changes requested):
Looks good overall, only major issue is not writing sample ToC.

tools/metadata_tools/process_metadata.py
duffh: Good catch

tools/metadata_tools/process_metadata.py
duffh: I need some clarification here.

ZackAllen (approved):

PR #1141, opened 2022-11-15
Run sample_sync on v.next 

duffh:
# Description

Ran the `sample_sync` script.

## Type of change

- [ ] Bug fix
- [ ] New sample implementation
- [ ] Sample viewer enhancement
- [x] Other enhancement


duffh:
There may be an issue with the `sample_sync` check as it is registering a change when running the github action but there is no change when I run the script locally.

ZackAllen (approved):

PR #1217, opened 2023-04-14
Removed web map from Open map URL sample

duffh:
# Description

Removed a web map from the Open map URL sample as this web map is no longer active. 

## Type of change

- Other enhancement

## Platforms tested on

- [x] WPF .NET 6
- [ ] WPF Framework
- [x] WinUI
- [x] MAUI WinUI
- [ ] MAUI Android
- [ ] MAUI iOS
- [ ] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes
- [ ] All changes work as expected on all affected platforms

### DIFF
--- a/src/MAUI/Maui.Samples/Samples/Map/OpenMapURL/OpenMapURL.xaml.cs
+++ b/src/MAUI/Maui.Samples/Samples/Map/OpenMapURL/OpenMapURL.xaml.cs
@@ -22,14 +22,12 @@ namespace ArcGIS.Samples.OpenMapURL
         // String array to hold urls to publicly available web maps.
         private string[] _itemURLs = {
             "https://arcgis.com/home/item.html?id=92ad152b9da94dee89b9e387dfe21acd",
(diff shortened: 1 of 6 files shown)


duffh:
The check that is failing is a result of the URL not being wrapped in `< >` braces. This is not a standard we have been following in our other readme files so I will merge with the check failing once this has been approved.

duffh:
The check that is failing is a result of the URL not being wrapped in `< >` braces. This is not a standard we have been following in our other readme files so I will merge with the check failing once this has been approved.

williambohrmann3 (approved):
Do we need to update MD034 (markdown rule) to allow this type of link in our readme's? This format with the [name](link) used in other samples (i.e. Query Feature Count and Extent) so it probably shouldn't be flagging it.

amweis226 (approved):

duffh:
> Do we need to update MD034 (markdown rule) to allow this type of link in our readme's? This format with the `[name](link)` is used in other samples (i.e. Query Feature Count and Extent) so it probably shouldn't be flagging it as an unsuccessful check.

I think the best approach is to remove it, I'll see if I can find out why it's not been an issue until just now.

PR #1228, opened 2023-04-20
StatisticalQuery UI enhancements

williambohrmann3:
# Description

- Updated the statistics to make them more readable.
- Click events to follow naming conventions.
- Used stack layouts for UI when possible.

## Type of change

- Other enhancement

## Platforms tested on

- [x] WPF .NET 6
- [x] WPF Framework
- [x] WinUI
- [x] MAUI WinUI
- [x] MAUI Android
- [x] MAUI iOS
- [x] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes
- [x] Screenshots are correct size and display in description tab (800 x 600 on Windows, 600 height mobile screenshots for MAUI)


duffh (changes requested):
Add full stops at the end of the comments in these files. We have been including these when we write new samples or update older ones. 

I'd include a dismiss button on WPF and WinUI for parity with the MAUI sample. 

Otherwise it looks good.

amweis226 (changes requested):
Agree with Hamish's comments.

In addition, if a search result turns up no statistics (zoom to a place with no markers and check the first box for example), an error throws (Error: Unexpected null value). Seems handled but in these cases we could either just return in the code if 0 results are returned or change the wording on the error so that popping up a MessageBox (or DisplayAlert in the case of MAUI) is more clearly intentional. My vote is for the former. 

Also (more of a nitpick but worth noting), the field names in results are all lower case. Typical GIS practice would usually capitalize at least the first word.

Otherwise, looks good to me.

williambohrmann3:
In the case that the result returns no statistics, calling `statQueryResult.First().Statistics` will throw an `ArgumentException`. So some exception handling is necessary. I thought throwing the exception ourselves would be safer. This way, we know for certain where the exception was thrown and why.

amweis226:
> In the case that the result returns no statistics, calling `statQueryResult.First().Statistics` will throw an `ArgumentException`. So some exception handling is necessary. I thought throwing the exception ourselves would be safer. This way, we know for certain where the exception was thrown and why.

Makes sense, I didn't have time to make sure we could return on null results. I think at least the wording could be changed to something like "0 results returned." or something similar. "Unexpected null value" implies it's an error as opposed to an expected behavior given that specific condition.

duffh (approved):

amweis226 (approved):

PR #1239, opened 2023-04-28
Add `.editorconfig`

williambohrmann3:
# Description

Including a `.editorconfig` file at the solution level will apply styling rules within the project. It will not overwrite IDE preferences, making it a good tool for styling. See the comments on the top of the file as to how to run the tool. We could run this command once before 

To help with styling consistently: Running dotnet-format on the entire solution, say once a week, and pushing any changes could be a way to automate styling via a GH Action. However, I think manually running dotnet-format, say before a program increment goes live would be sufficient.

## Type of change

- Other enhancement

## Checklist

- [x] Self-review of changes


williambohrmann3:
The CodeMaid "Cleanup All Code" options is dramatically slower (at least on my build) than `dotnet-format` command.

PR #1258, opened 2023-06-23
Vector tiled layer custom style cache bug fix and code cleanup

williambohrmann3:
# Description

- await for cache to load when using offline custom style so layer actually displays
- one array for portal items
- determine if a user selected a custom style using style name (no longer using arbitrary index)
- MAUI: replace DisplayActionSheet with Picker

## Type of change

- Bug fix
- Sample code cleanup

## Platforms tested on

- [x] WPF .NET 6
- [x] WPF Framework
- [x] WinUI
- [x] MAUI WinUI
- [ ] MAUI Android
- [x] MAUI iOS
- [x] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes

### DIFF
--- a/src/MAUI/Maui.Samples/Samples/Layers/AddVectorTiledLayerFromCustomStyle/AddVectorTiledLayerFromCustomStyle.xaml.cs
+++ b/src/MAUI/Maui.Samples/Samples/Layers/AddVectorTiledLayerFromCustomStyle/AddVectorTiledLayerFromCustomStyle.xaml.cs
@@ -95,37 +95,37 @@ namespace ArcGIS.Samples.AddVectorTiledLayerFromCustomStyle
             }
         }
 
-        private void StyleChooser_SelectedIndexChanged(object sender, EventArgs e)
-        {
-            // Get the style name and index of the selected item.
-            _ = ChangeStyle(StyleChooser.SelectedIndex, StyleChooser.SelectedItem.ToString());
-        }
-
-        private async Task ChangeStyle(int styleIndex, string styleName)
+        private async void StyleChooser_SelectedIndexChanged(object sender, EventArgs e)
         {
             try
             {
-                // Check if the user selected an online or offline custom style.
-                // Create a new basemap with the appropriate style.
-                if (styleName.Contains("Offline"))
-                {
-                    // Determine which cache to use based on if the style selected is light or dark.
-                    ItemResourceCache cache = styleName.Contains("Light") ? _lightStyleResourceCache : _darkStyleResourceCache;
-
-                    MyMapView.Map.Basemap = new Basemap(new ArcGISVectorTiledLayer(new VectorTileCache(_localVectorPackagePath), cache));
-                    await MyMapView.SetViewpointAsync(_dodgeCityViewpoint);
-                    await cache.LoadAsync();
-                }
-                else
-                {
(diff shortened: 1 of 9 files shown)


src/MAUI/Maui.Samples/Samples/Layers/AddVectorTiledLayerFromCustomStyle/AddVectorTiledLayerFromCustomStyle.xaml.cs
```diff
@@ -94,37 +95,31 @@ private async Task Initialize()
...
-        private void ButtonClick(object sender, EventArgs e)
+        private void StyleChooser_SelectedIndexChanged(object sender, EventArgs e)
         {
-            _ = StyleChooser();
+            // Get the style name and index of the selected item.
+            _ = ChangeStyle(StyleChooser.SelectedIndex, StyleChooser.SelectedItem.ToString());
         }
 
-        private async Task StyleChooser()
+        private async Task ChangeStyle(int styleIndex, string styleName)
```
mstefarov, line 104: I would rename this to ChangeStyleAsync.  This is a [general recommendation](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/async-scenarios#important-info-and-advice) by the .NET team:
> Add "Async" as the suffix of every async method name you write.

src/MAUI/Maui.Samples/Samples/Layers/AddVectorTiledLayerFromCustomStyle/AddVectorTiledLayerFromCustomStyle.xaml.cs
```diff
@@ -94,37 +95,31 @@ private async Task Initialize()
             }
         }
 
-        private void ButtonClick(object sender, EventArgs e)
+        private void StyleChooser_SelectedIndexChanged(object sender, EventArgs e)
         {
-            _ = StyleChooser();
+            // Get the style name and index of the selected item.
+            _ = ChangeStyle(StyleChooser.SelectedIndex, StyleChooser.SelectedItem.ToString());
         }
```
mstefarov, line 102: I like this change!  The separation between app logic ("changing style") and GUI logic (interacting with specific controls) is better now.

Optional: I think ChangeStyle can be even cleaner if you moved try/catch/report out of it.  Something like...

```cs
private async void StyleChooser_SelectedIndexChanged(object sender, EventArgs e)
{
    try
    {
        // Get the style name and index of the selected item.
        await ChangeStyle(StyleChooser.SelectedIndex, StyleChooser.SelectedItem.ToString());
    }
    catch (Exception ex)
    {
        // Report exceptions.
        await Application.Current.MainPage.DisplayAlert("Error", ex.Message, "OK");
    }
}
```

This makes ChangeStyle into a simpler, more reusable method that only does what it says (changes style) without extra side-effects.

mstefarov (approved):
I like the changes!

src/MAUI/Maui.Samples/Samples/Layers/AddVectorTiledLayerFromCustomStyle/AddVectorTiledLayerFromCustomStyle.xaml.cs
williambohrmann3: Good to know!

duffh (approved):

PR #1261, opened 2023-06-28
Explicit cleanup upon closing samples

williambohrmann3:
# Description

- Ensures location services are disabled
- Nullifies maps and scenes
- Removes life cycle workaround for iOS/MacCatalyst - event works fine now

## Type of change

- Sample viewer enhancement

## Platforms tested on

- [x] WPF .NET 6
- [x] WPF Framework
- [x] WinUI
- [x] MAUI WinUI
- [ ] MAUI Android
- [x] MAUI iOS
- [x] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes

### DIFF
--- a/src/MAUI/Maui.Samples/SamplePage.xaml.cs
+++ b/src/MAUI/Maui.Samples/SamplePage.xaml.cs
@@ -17,6 +17,7 @@
 
 using ArcGIS.Samples.Managers;
 using ArcGIS.Samples.Shared.Models;
+using Esri.ArcGISRuntime.Maui;
 using Microsoft.Maui.Platform;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
@@ -44,12 +45,7 @@ namespace ArcGIS
 
         public SamplePage(ContentPage sample, SampleInfo sampleInfo) : this()
         {
-            this.NavigatedFrom += NavigatedFromEvent;
-
-#if IOS || MACCATALYST
-            // iOS / MacCatalyst lifecycle works differently, so we need to use the main page changing instead of the NavigatedFrom event for this.
-            Application.Current.MainPage.PropertyChanged += MainPagePropertyChanged;
-#endif
+            NavigatedFrom += SamplePage_NavigatedFrom;
 
             // Set the sample variable.
             _sample = sample;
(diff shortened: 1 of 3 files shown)


src/WinUI/ArcGIS.WinUI.Viewer/SamplePage.xaml.cs
```diff
@@ -37,6 +39,8 @@ public SamplePage()
             // Load and show the sample.
             SampleContainer.Content = SampleManager.Current.SampleToControl(SampleManager.Current.SelectedSample);
 
+            (SampleContainer.Content as FrameworkElement).Unloaded += SamplePage_Unloaded;
```
dotMorten, line 42: Event is already being used for ContentPage.Unloaded, and you're using the same eventhandler. Is this line even needed? The Unloaded handler should already be called

williambohrmann3:
This PR and the [memory leak PR](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/pull/1262) both touch `SamplePage.xaml.cs`. I can reapply my MAUI-specific changes when the memory leak PR is complete as to avoid a merge conflict.

dotMorten (approved):

src/MAUI/Maui.Samples/SamplePage.xaml.cs
```diff
@@ -112,29 +108,23 @@ private async void LoadSampleData(SampleInfo sampleInfo)
...
-        private void NavigatedFromEvent(object sender, NavigatedFromEventArgs e)
-        {
-            TryDispose();
-        }
-
-        private void TryDispose()
+        private void SamplePage_NavigatedFrom(object sender, NavigatedFromEventArgs e)
         {
             // Check that the navigation is backward from the sample and not forward into another page in the sample.
             if (!Application.Current.MainPage.Navigation.NavigationStack.OfType<SamplePage>().Any())
```
mstefarov, line 114: I believe this will always be false, because current page is still in the NavigationStack when NavigatedFrom event fires. Try setting a breakpoint inside this method to see if it's actually reached.

williambohrmann3: It's making it inside the conditional when I place a breakpoint at the for loop. It's having expected behavior - I tested the ArcGIS token challenge sample which has an additional page. And tested a sample without an additional page.

src/WPF/WPF.Viewer/MainWindow.xaml.cs
```diff
@@ -227,6 +229,39 @@ private async Task SelectSample(SampleInfo selectedSample)
...
+        {
+            if (root != null)
+            {
+                if (root is T t)
+                    yield return t;
+                else if (root is UIElement it)
+                {
+                    for (int i = 0; i < VisualTreeHelper.GetChildrenCount(it); i++)
+                    {
+                        var e = VisualTreeHelper.GetChild(it, 0);
```
mstefarov, line 257: Did you mean to get the `i`th child, instead of always getting the first child?
```suggestion
                        var e = VisualTreeHelper.GetChild(it, i);
```

src/WinUI/ArcGIS.WinUI.Viewer/SamplePage.xaml.cs
```diff
@@ -87,6 +89,36 @@ private void SamplePage_Loaded(object sender, RoutedEventArgs e)
...
+        {
+            if (root is not null)
+            {
+                if (root is T t)
+                    yield return t;
+                else if (root is UIElement it)
+                {
+                    for (int i = 0; i < VisualTreeHelper.GetChildrenCount(it); i++)
+                    {
+                        var e = VisualTreeHelper.GetChild(it, 0);
```
mstefarov, line 116: Same here
```suggestion
                        var e = VisualTreeHelper.GetChild(it, i);
```

src/MAUI/Maui.Samples/SamplePage.xaml.cs
mstefarov: Hmm you're right, I just retested and it works.  False alarm!  I'm not sure what I was doing wrong before, sorry about that.

src/MAUI/Maui.Samples/SamplePage.xaml.cs
```diff
@@ -112,29 +108,23 @@ private async void LoadSampleData(SampleInfo sampleInfo)
...
-            }
-        }
-
-        private void NavigatedFromEvent(object sender, NavigatedFromEventArgs e)
-        {
-            TryDispose();
-        }
-
-        private void TryDispose()
+        private void SamplePage_NavigatedFrom(object sender, NavigatedFromEventArgs e)
```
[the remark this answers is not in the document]
williambohrmann3: In my Visual Studio environment, whenever subscribing to an event, it autogenerates the method with a `ClassName_EventName` name. I think we should follow that convention for readability; we name event handlers this way in our samples.

mstefarov (approved):

PR #1274, opened 2023-08-01
Update samples using sketch editor to use geometry editor

duffh:
# Description

As sketch editor is being replaced by geometry editor we are replacing sketch editor usage in samples with geometry editor.

## Type of change

- Other enhancement

## Platforms tested on

- [x] WPF .NET 6
- [x] WPF Framework
- [x] WinUI
- [x] MAUI WinUI
- [x] MAUI Android
- [ ] MAUI iOS
- [ ] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes


jnery (changes requested):
Thanks for getting this started @duffh. I just have few suggestions.

I think how you have it now works too. Only suggesting:
- projection necessary
- nameof() syntax
- unnecessary TaskCanceledException catch
- unnecessary Stop when Start is about to be called again
- if double tapped gesture is needed, we can still call Stop there

src/WPF/WPF.Viewer/Samples/NetworkAnalysis/FindServiceArea/FindServiceArea.xaml.cs
```diff
@@ -50,46 +53,22 @@ private void Initialize()
...
         {
             try
             {
-                // Let the user tap on the map view using the point sketch mode.
-                SketchCreationMode creationMode = SketchCreationMode.Point;
-                Geometry geometry = await MyMapView.SketchEditor.StartAsync(creationMode, false);
-
-                // Symbology for a facility.
-                PictureMarkerSymbol facilitySymbol = new PictureMarkerSymbol(new Uri("https://static.arcgis.com/images/Symbols/SafetyHealth/Hospital.png"))
+                if (MyMapView.GeometryEditor.IsStarted)
```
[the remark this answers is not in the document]
jnery: You can ignore this feedback. Your changes look consistent in that you handle completed geometry in `GeometryEditor.PropertyChanged` handler so the Stop() here should trigger that event.

This suggestion was in case you wanted to keep the following syntax:
```CSharp
var geometry  = await StartAsync(..)
// do something with geometry
```
to do that, you can use `TaskCompletionSource<Geometry> _tcs` that you set result in the event handler so it becomes
```CSharp
var geometry = await _tcs.Task; // will return once TrySetResult/Exception/Canceled are called.
```

I think how you translated legacy code is just as valid, clean and consistent. Marking this as resolved.

jnery (approved):
Looks good 👍  thanks for staying on top of deprecation warnings :)

src/MAUI/Maui.Samples/Samples/NetworkAnalysis/FindServiceArea/FindServiceArea.xaml.cs
```diff
@@ -99,14 +78,50 @@ private async void PlaceFacilityButton_Click(object sender, EventArgs e)
             }
         }
 
+
+        private void GeometryEditor_PropertyChanged(object sender, System.ComponentModel.PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == nameof(GeometryEditor.Geometry))
+            {
+                if (_geometryType == GeometryType.Point)
```
williambohrmann3, line 86: Can we combine these two if checks with && ?

src/MAUI/Maui.Samples/Samples/Data/GeodatabaseTransactions/GeodatabaseTransactions.xaml.cs
```diff
@@ -105,53 +109,53 @@ private async Task GetLocalGeodatabase()
...
+                    // Store a reference to the Marine table.
                     if (table.TableName.ToLower().Contains("marine"))
                     {
                         _marineTable = table;
                     }
 
-                    // Create a new feature layer to show the table in the map
+                    // Create a new feature layer to show the table in the map.
                     FeatureLayer layer = new FeatureLayer(table);
                     Microsoft.Maui.ApplicationModel.MainThread.BeginInvokeOnMainThread(() => MyMapView.Map.OperationalLayers.Add(layer));
```
williambohrmann3, line 160: On iOS sim, this line of code is throwing a `NullReferenceException`:
image
I can also repro this on v.next.

src/MAUI/Maui.Samples/Samples/Data/GeodatabaseTransactions/GeodatabaseTransactions.xaml.cs
```diff
@@ -105,53 +109,53 @@ private async Task GetLocalGeodatabase()
...
                     _localGeodatabase = await generateGdbJob.GetResultAsync();
                 }
             }
             catch (Exception ex)
             {
-                // Show a message for the exception encountered
+                // Show a message for the exception encountered.
                 Microsoft.Maui.ApplicationModel.MainThread.BeginInvokeOnMainThread(() =>
                 {
                     Application.Current.MainPage.DisplayAlert("Generate Geodatabase", "Unable to create offline database: " + ex.Message, "OK");
```
williambohrmann3, line 128: We are catching this exception on MacCatalyst, sample is unusable. Can also repro on v.next.

src/MAUI/Maui.Samples/Samples/Data/GeodatabaseTransactions/GeodatabaseTransactions.xaml.cs
jnery: So this fails even without this PR changes? Are you able to check which property is null on this line? Is it MapView, Map or ApplicationModel or MainThread?

williambohrmann3: Yep, this issue pre-dates this PR. To repro, launch sample and return to category page. MapView.Map is set as null [OnNavigatedFrom](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/blob/v.next/src/MAUI/Maui.Samples/SamplePage.xaml.cs#L114) so before my breakpoint on line 160 is hit. I will see if I can repro on other MAUI platforms.

src/MAUI/Maui.Samples/Samples/Data/GeodatabaseTransactions/GeodatabaseTransactions.xaml.cs
williambohrmann3: Screenshot 2023-08-02 at 1 47 15 PM

src/MAUI/Maui.Samples/Samples/Data/GeodatabaseTransactions/GeodatabaseTransactions.xaml.cs
jnery: If samples is setting Map to null, then we need a null check here. May be a different PR, sounds like it can happen else where Map is accessed on an invoke after you navigated away from a sample.

williambohrmann3: Can confirm using null operator (MyMapView.Map?.OperationalLayers) on line 160 prevents this exception on iOS.

src/MAUI/Maui.Samples/Samples/Data/GeodatabaseTransactions/GeodatabaseTransactions.xaml.cs
duffh: Fixed with [commit](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/pull/1274/commits/defcbce79b2a9bcf67a1dcef7a08e399ae4a66b8).

williambohrmann3 (approved):
glad you were able to fix the pre-existing issues!

PR #1283, opened 2023-08-11
Fix issue with Geodatabase transactions sample when navigating away

duffh:
# Description

A missing null check resulted in a thrown exception when navigating away from the Geodatabase transactions sample.

## Type of change

- Bug fix

## Platforms tested on

- [x] WPF .NET 6
- [x] WPF Framework
- [x] WinUI

## Checklist

- [x] Self-review of changes


williambohrmann3 (approved):
Thanks! tested both sample implementations.

williambohrmann3:
Quite rare, have only been able to repro this once after about 10 attempts:
getting these errors after changing samples then returning to `GeodatabaseTransactions`:
devenv_EZfguTkMOa
Might be worth looking into but I don't think this is related to your PR. Very odd I only could repro this once.

PR #1308, opened 2023-10-10
GitHub PR automation: account for already open PR

williambohrmann3:
This is a workaround for job failures caused by a PR already existing for `v.next <- main`. An unhandled return code 1 was to blame. We want to account for this so the job is a success in this common case.


williambohrmann3:
See [test50](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/actions/runs/6474435889/job/17579342378#step:3:25) which demos this case.

williambohrmann3:
Closing - this logic has been added to `wbohr/cherry-pick-from-main`

PR #1313, opened 2023-10-30
PR automation: cherry-pick from main

williambohrmann3:
# Description

An issue with the current `Open_PR` workflow is dealing with merge conflicts. We have a few minor but intentional differences between main and `v.next`, for example, `Directory.Packages.props`.  This results in a less streamlined process of keeping `v.next` synced with bug fixes and enhancements that go directly into `main`. 

The solution proposed is to change the head of the generated PR. Currently, that branch is `main`. Instead, we can branch off `v.next` and cherry-pick the most recent commit of `main`. This way, the head has the same commit history as `v.next` but with the additional cherry-picked commit. This will reduce the risk of merge conflicts and reduce PR scope. Multiple "sync" PRs may be opened at once since the modified job generates a new branch per commit to `main`.

Another issue with the current workflow is if a PR fails to open, the job fails. I've added some output which writes if a diff is detected and if opening a PR is successful. The job will no longer fail if the PR cannot be opened. An advantage of this modified workflow is no more opening identical PRs. The branch name will follow a convention of sync/commit-hash, where the commit hash is the most recent commit from main.

In the checkout code step, I've added parameters to checkout a shallow clone of `main` at depth 2. This truncates the git history and doesn't download history of branches unrelated to the job. The same kind of clone for `v.next` is made in the following step. 

For more, see:
- [Successful run](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/actions/runs/6699836972/job/18204699060) of this job.
- `wbohr/cherry-pick-from-main-test` commit history and triggered test runs.

## Type of change

- Automation

## Checklist

- [x] Self-review of changes

### DIFF
--- a/.github/workflows/Open_PR.yml
+++ b/.github/workflows/Open_PR.yml
@@ -1,5 +1,6 @@
 name: Sync v.next with main
 
+# PRs are completed with squash and merge option - this commit will be cherry-picked for v.next.
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


cgutierrez-gis (approved):
looks good!

duffh (approved):

PR #1317, opened 2023-11-07
Define rows/columns of sample container grid upon device orientation change

williambohrmann3:
# Description

On iPhone, the control panel becomes unusable when changing from portrait to landscape device orientation on many samples. The control panel is currently drawn in a column of width `auto`. There is a difference in the way `auto` is being determined on Androids and iPhones for this column width.

My PR has all samples take advantage of the `EsriSampleContainer` style of `Grid`. I've added a `VisualStateManager` so that way when in landscape orientation, use `2*,*`for column definitions. That way, the control panel has enough room to render controls, allowing samples to be usable in landscape orientation on iOS.

We previously dereferenced the `EsriSampleContainer` due to a memory leak when accessing the `DynamicResource`. This issue has been since fixed by the .NET MAUI team via this [PR](https://github.com/dotnet/maui/pull/16145).

Also, this PR adds a `MaximumWidthRequest` to the control panel style when in portrait mode. Since the row height is `auto` where the control panel rests, it will minimize the space of the control panel as to maximize the real estate for the `GeoView`. If the control panel has a height > 300 on phone and > 500 on tablet, the control panel will use this `MaximumWidthRequest`.

## Type of change

- Bug fix

## Platforms tested on

- [x] MAUI WinUI
- [x] MAUI Android
- [x] MAUI iOS
- [x] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes

### DIFF
--- a/src/MAUI/Maui.Samples/Resources/Styles/Styles.xaml
+++ b/src/MAUI/Maui.Samples/Resources/Styles/Styles.xaml
@@ -7,8 +7,39 @@
 
     <!--  Esri samples helper styles  -->
     <Style x:Key="EsriSampleContainer" TargetType="Grid">
-        <Setter Property="Grid.RowDefinitions" Value="Auto,*" />
-        <Setter Property="Grid.ColumnDefinitions" Value="*,Auto" />
+        <Setter Property="VisualStateManager.VisualStateGroups">
+            <VisualStateGroupList>
+                <VisualStateGroup>
+                    <VisualState x:Name="Vertical">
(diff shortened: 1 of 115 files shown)


williambohrmann3:
Note: targeting `v.next` as to take advantage of latest fixes in .NET 8.

williambohrmann3:
before:
before
after:
after

williambohrmann3:
before:
before

after:
after

duffh (approved):

PR #1391, opened 2024-01-30
Add clustering feature reduction to a point feature layer fixes

williambohrmann3:
# Description

- Replaces sliders with pickers to avoid flickering
- Updates UI
- Fixes a crash caused by drawing clusters before map finishes loading

## Type of change

- Bug fix
- Other enhancement

## Platforms tested on

- [x] WPF .NET 8
- [x] WPF Framework
- [x] WinUI
- [x] MAUI WinUI
- [x] MAUI Android
- [x] MAUI iOS
- [x] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes


src/MAUI/Maui.Samples/Samples/Layers/AddClusteringFeatureReductionToAPointFeatureLayer/AddClusteringFeatureReductionToAPointFeatureLayer.xaml.cs
```diff
@@ -137,14 +146,18 @@ private void DisplayLabelsCheckBox_CheckedChanged(object sender, CheckedChangedE
             {
                 _clusteringFeatureReduction.LabelDefinitions.Clear();
             }
-            
         }
 
-        private void UpdateClusteringProperties(object sender, EventArgs e)
+        private void ClusterRadiusPicker_SelectedIndexChanged(object sender, EventArgs e)
```
duffh, line 151: Do we need two event handlers here? 

Previously calling `UpdateClusteringProperties` covered both bases.

Also I think we should maintain our existing format of:

```
if(_layer != null)
{
    // do something
}
```

I don't think bracketless if statements are used in this format anywhere else.

src/WinUI/ArcGIS.WinUI.Viewer/Samples/Layers/AddClusteringFeatureReductionToAPointFeatureLayer/AddClusteringFeatureReductionToAPointFeatureLayer.jpg
duffh, line 1: Screenshot with orange checkbox background looks out of place.

src/MAUI/Maui.Samples/Samples/Layers/AddClusteringFeatureReductionToAPointFeatureLayer/AddClusteringFeatureReductionToAPointFeatureLayer.xaml.cs
williambohrmann3: I kept running into null reference exceptions (specifically WPF) when I modified `UpdateClusteringProperties` to account for the swapping of the sliders with pickers. The issue is `MaxScalePicker.SelectedItem` is null when setting `ClusterRadiusPicker.SelectedItem`. Technically we might be able to null check on the picker selected items before trying to use them, however that introduces extra logic that can be avoided by having two event handlers. Furthermore, using two events allows each control to handle their own property setting separately, avoiding unnecessary property assignments. Happy to use curly braces for the if statements.

src/WinUI/ArcGIS.WinUI.Viewer/Samples/Layers/AddClusteringFeatureReductionToAPointFeatureLayer/AddClusteringFeatureReductionToAPointFeatureLayer.jpg
williambohrmann3: > Screenshot with orange checkbox background looks out of place.

WinUI uses the system theme color by default, which is orange for me and it looks like it is blue for you based on the previous screenshot. We want to display labels in the screenshot, so the checkbox background is filled with my system color.

duffh: I think we should be consistent with our system color when taking capturing images of our samples as it looks strange having a mix of different colors in our screenshots, particularly when scrolling through the samples categories.

duffh (approved):

PR #1398, opened 2024-02-05
Replace .NETFramework.csproj in samplegen.py

williambohrmann3:
# Description

Don't use the .NETFramework .csproj which doesn't exist anymore in our sample generator script. Use the existing .csproj instead. Found this when creating `ValidateUtilityNetworkTopology` sample and subsequently tested my fix.

## Type of change

- Bug fix


duffh (approved):
Good catch

PR #1447, opened 2024-05-17
samplegen: remove path parameter

williambohrmann3:
# Description

When running the sample generator script, determine the path using the `os.path` functions instead of relying on user. This should make the tool easier, resulting in a better development experience.

## Type of change

- Other enhancement


duffh (approved):
👍

PR #1457, opened 2024-06-18
MAUI viewer: search optimization

williambohrmann3:
# Description

Populates a dictionary with the key being the sample and the value being the keywords. Replaces for loop with dictionary lookups.

## Type of change

- Sample viewer enhancement

## Platforms tested on

- [x] MAUI WinUI
- [x] MAUI Android
- [ ] MAUI iOS
- [ ] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes

### DIFF
--- a/src/MAUI/Maui.Samples/ViewModels/SearchViewModel.cs
+++ b/src/MAUI/Maui.Samples/ViewModels/SearchViewModel.cs
@@ -15,6 +15,11 @@ namespace ArcGIS.ViewModels
         private Dictionary<SampleInfo, (string[] nameKeywords, string[] categoryKeywords, string[] descriptionKeywords, string[] tagsKeywords)>
             _sampleKeywords = new();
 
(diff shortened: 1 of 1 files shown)


williambohrmann3:
It also might be a good idea to add `if (sampleResults.Count >= 15) break;` within the `GetMatches()` loop to focus the results on high quality matches and for a (very) minor performance bump. (right after we add the results to `samplesResult`) Thoughts?

mstefarov (approved):
Good improvements!

williambohrmann3:
just committed some further search optimizations, credits to @mstefarov 

> - Uses regex generator (faster) instead of creating a new Regex every time.
> - Use a static HashMap for common words instead of re-creating the array over and over.
> - Do trim as part of split using StringSplitOptions.TrimEntries
> - Avoid an extra ToList() call

PR #1557, opened 2025-02-26
Samples consolidation: `Configure electronic navigational charts` 

williambohrmann3:
# Description

Adds `Configure electronic navigational charts` sample to `Layers` category. Consolidates all `Hydrography` samples.

## Type of change

- New sample implementation

## Platforms tested on

- [x] WPF .NET 8
- [x] WPF Framework
- [x] WinUI
- [x] MAUI WinUI
- [x] MAUI Android
- [ ] MAUI iOS
- [ ] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes
- [x] All changes work as expected on all affected platforms
- [x] There are no warnings related to changes
- [x] Code is commented and follows .NET conventions and standards
- [x] Codemaid and XAML styler extensions have been run on every changed file
- [x] No unrelated changes have been made to any other code or project files
- [x] Screenshots are correct size and display in description tab (800 x 600 on Windows, MAUI mobile platforms use the MAUI Windows screenshot)

PR #1565, opened 2025-03-17
Fix redirect links

williambohrmann3:
# Description

Undoes redirect link removal which unintentionally occurred via #1556 
Undoes wrong redirect links addition which unintentionally occurred via #1557 
Note that we should update our sample sync script to not remove extra redirect links for certain samples

## Type of change

- Bug fix


duffh (approved):
LGTM

PR #1583, opened 2025-04-05
uwp: Fix use of embedded marker symbols

mstefarov:
# Description
Another fix for UWP samples from certification:

- The old fashioned "GetManifestResourceStream" method only works in Debug builds. The recommended pattern in UWP is to use "Content" build action and ms-appx URIs. This avoids a variety of crashes/exceptions when using the sample viewer in Release mode, and shows the best practice to our users.
- Additional fix for _Mobile Map (Search and Rescue)_ only: Map thumbnails did not work with ItemToImageSourceConverter  because LocalItems only have a Thumbnail and not a ThumbnailUri. Conversion has to be done asynchronously so an IValueConverter will not work. After some brainstorming I came up with an attached-property converter, which saves us from having to create MapViewModels and ItemViewModels and all the related plumbing. Perhaps our users will find this trick useful!

## Type of change
- Bug fix

## Platforms tested on
- [x] UWP

## Checklist

- [x] Self-review of changes


duffh (approved):
👍

PR #1619, opened 2025-06-04
Tools: don't use `latest` in redirect paths

williambohrmann3:
Removes `latest` from redirect links in sample generator and sample metadata scripts


cgutierrez-gis (approved):
lgtm

PR #1638, opened 2025-07-23
Adds missing config options

williambohrmann3:
I've had to manually enable several config options, so it makes sense to have these enabled.
- adds missing config options to test on release config and arm64 devices
- fixes csproj spacing issue


src/ArcGIS.AllSampleViewers.sln
mstefarov, line 1: The mappings in this all-viewer solution seem off.  ARM64 maps to x86?  x64 maps to AnyCPU?  I would look over it in Configuration Manager and make sure everything is consistent.

mstefarov (changes requested):

mstefarov (approved):
Looks good, works for me locally too 👍

PR #1653, opened 2025-08-28
README update - remove deprecated platforms

williambohrmann3:
- removes mention of UWP and .NET Framework
- .NET MAUI sln file is listed twice, so reducing to one


praveenaak (approved):
LGTM

PR #1670, opened 2025-09-26
Ian/deprecation fixes main

imalcolm1:
# Description

Fixes for deprecation warnings related to the following:
- Use of Application.Current.MainPage in the MAUI offline routing sample
- Setting VerticalOptions or HorizontalOptions to FillAndExpand or CenterAndExpand options in MAUI XAML files
- Use of deprecated Frame element in MAUI

## Type of change

- Bug fix

## Platforms tested on

- [x] WPF .NET 8
- [x] WinUI
- [x] MAUI WinUI
- [x] MAUI Android
- [x] MAUI iOS
- [ ] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes

### DIFF
--- a/src/MAUI/Maui.Samples/Samples/Analysis/ViewshedGeoElement/ViewshedGeoElement.xaml
+++ b/src/MAUI/Maui.Samples/Samples/Analysis/ViewshedGeoElement/ViewshedGeoElement.xaml
@@ -3,7 +3,7 @@
              xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
              xmlns:esriUI="clr-namespace:Esri.ArcGISRuntime.Maui;assembly=Esri.ArcGISRuntime.Maui">
-    <Grid RowDefinitions="Auto,*" Style="{DynamicResource EsriSampleContainer}">
+    <Grid Style="{DynamicResource EsriSampleContainer}">
         <esriUI:SceneView x:Name="MySceneView" Style="{DynamicResource EsriSampleGeoView}" />
         <Border Style="{DynamicResource EsriSampleControlPanel}">
             <Label Text="Tap to set a destination for the vehicle." FontSize="16" />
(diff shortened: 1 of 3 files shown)


src/MAUI/Maui.Samples/Samples/Analysis/ViewshedGeoElement/ViewshedGeoElement.xaml
```diff
@@ -3,8 +3,8 @@
              xmlns="http://schemas.microsoft.com/dotnet/2021/maui"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
              xmlns:esriUI="clr-namespace:Esri.ArcGISRuntime.Maui;assembly=Esri.ArcGISRuntime.Maui">
-    <StackLayout Orientation="Vertical">
-        <Label Text="Tap to set a destination for the vehicle." VerticalOptions="Start" />
-        <esriUI:SceneView x:Name="MySceneView" VerticalOptions="FillAndExpand" />
-    </StackLayout>
+    <Grid RowDefinitions="Auto,*">
```
williambohrmann3, line 6: There are multiple places in this file where you are replicating the UI of the old sample implementation, which is valid. However, since you are touching the UIs of these files, I do think you should follow our standard UI design pattern. That involves using the Border, Grid, and GeoView Styles as defined in https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/blob/main/src/MAUI/Maui.Samples/Resources/Styles/Styles.xaml#L7-L122. By following our UI design pattern, we can reduce the variability in sample UIs which should reduce the risk of breaking changes in future versions of .NET. In the future the hope is we can fix the UI in one place rather than in individual samples as you had to do with this PR.

williambohrmann3 (commented):
Remove the row definitions from the grid of those 3 samples which use the SampleContainerStyle. Visual state setters in the Style handle that. Otherwise, looks good.

williambohrmann3 (approved):

PR #1671, opened 2025-09-27
Ian/UI fixes main

imalcolm1:
# Description

A set of minor UI fixes for the following bugs.

#### Author map (MAUI, WPF)
On MAUI android, the user is unable to click the save button as it is out of frame and cannot be scrolled to
- This is true in either landscape or portrait orientation
- Occurs on both physical tablet and pixel phone simulator

In WPF, the choose layers ListBox is not showing despite appearing in the Visual Tree debug pane

#### Database branch versioning (MAUI)
There is no error message when trying to create a branch with the same name as a preexisting branch

#### Route around barriers (All)
There is a "Calculating" dialog that never shows

#### PKI Auth (WinUI)
The code attempts to connect before actually selecting the certificate. Removing this code seems to fix the issue, so it should be a very quick fix

## Type of change

- Bug fix

## Platforms tested on

- [x] WPF .NET 8
- [x] WinUI
- [x] MAUI WinUI
- [x] MAUI Android
- [ ] MAUI iOS
- [ ] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes

### DIFF
--- a/src/MAUI/Maui.Samples/Samples/Data/EditBranchVersioning/EditBranchVersioning.xaml.cs
+++ b/src/MAUI/Maui.Samples/Samples/Data/EditBranchVersioning/EditBranchVersioning.xaml.cs
@@ -178,7 +178,7 @@ namespace ArcGIS.Samples.EditBranchVersioning
 
         private void ShowAlert(string alertText, string titleText = "Alert")
         {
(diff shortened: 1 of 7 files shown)


williambohrmann3 (approved):
test MAUI Windows; LGTM
very good finds!

PR #1792, opened 2026-04-20
Merge main 300.0.0

imalcolm1:
# Description

Merge main into v.next to prepare for release. No actual changes, just lets git know the changes are already reconciled.

Edit: I forgot about the program_increment.py script. I think this still may be worthwhile to do after main gets updated. Will require a second merge, but that will be squashed.


imalcolm1:
We are going to delete and recreate v.next instead

PR #1807, opened 2026-05-19
ui: Lazy-load samples through generated catalog

mandel-capuchin:
# Description

Add a Roslyn source generator that emits an ISampleCatalogProvider with sample metadata and sample factories at build time. The generated provider lets startup build the category shell and Featured list without reflecting over every sample type or constructing every SampleInfo up front. This reduces launch work and helps avoid the iOS watchdog as the sample set grows.

Generated catalog shape:

```csharp
    public sealed class GeneratedSampleCatalog : ISampleCatalogProvider
    {
        public IReadOnlyList<string> GetCategories() => Categories;
        public IReadOnlyList<string> GetSampleNames() => SampleNames;

        public SampleInfo CreateSampleInfo(string formalName)
        {
            switch (formalName)
            {
                case "DisplayMap":
                    return new SampleInfo(
                        formalName: "DisplayMap",
                        sampleName: "Display map",
                        category: "Map",
                        sampleType: typeof(DisplayMap));
                default:
                    throw new ArgumentException(...);
            }
        }

        public object CreateSample(string formalName)
        {
            switch (formalName)
            {
                case "DisplayMap":
                    return new DisplayMap();
                default:
                    throw new ArgumentException(...);
            }
        }
    }
```

Old startup flow:

```text
    AppShell / flyout start
        |
        v
    SampleManager.Initialize()
        |
        v
    Assembly.GetTypes()
        |
        v
    Read attributes and create every SampleInfo
        |
        v
    Build full tree, Featured, Favorites
        |
        v
    First page can render
```

New startup flow:

```text
    AppShell / flyout start
        |
        v
    SampleManager.Initialize()
        |
        v
    GeneratedSampleCatalog via partial hook
        |
        v
    Load sample names and category names only
        |
        v
    Build category shell and materialize Featured
        |
        v
    First page can render
        |
        v
    Category, search, offline data, and samples load on demand
```

Route SampleManager through the catalog provider API for sample lookup, category loading, all-sample materialization, and sample creation. Search, offline data, and full sample lists still materialize on demand when those flows need them instead of during boot.

Keep ReflectionSampleCatalogProvider as a plan B for consumers or targets that do not wire in the generator. The fallback preserves the old reflection-based discovery behavior behind the same ISampleCatalogProvider contract while normal app builds use the generated catalog and avoid startup reflection.

## Type of change

- Sample viewer enhancement

## Platforms tested on

- [ ] WPF .NET 8
- [ ] WinUI
- [ ] MAUI WinUI
- [ ] MAUI Android
- [x] MAUI iOS
- [ ] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes

### DIFF
--- a/src/ArcGIS.AllSampleViewers.sln
+++ b/src/ArcGIS.AllSampleViewers.sln
@@ -17,6 +17,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ArcGIS.Samples.Maui", "MAUI
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ArcGIS.Samples.Shared", "Samples.Shared\ArcGIS.Samples.Shared.shproj", "{0AD85845-0BA1-4DC9-8C32-E6FD7632E32D}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.CatalogGenerator", "Samples.CatalogGenerator\Samples.CatalogGenerator.csproj", "{BBFBE218-B123-9B87-6B83-C46602227A09}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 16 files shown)


imalcolm1 (approved):
Looks good, loading is much faster everywhere now!

PR #1813, opened 2026-06-23
new sample: update basemap for contrast changes

praveenaak:
# Description

New Sample: Update basemap for contrast changes in the new Accessibility category

## Type of change

- New sample implementation

## Platforms tested on

- [x] WPF .NET 8
- [x] WinUI
- [x] MAUI WinUI
- [x] MAUI Android
- [x] MAUI iOS
- [x] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes


src/WPF/WPF.Viewer/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
imalcolm1, line 1: WPF sometimes gets into a bad state when switching between high contrast themes. The map never loads and only shows the background grid until clicking the manual toggle and back.

src/MAUI/Maui.Samples/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
```diff
@@ -0,0 +1,487 @@
...
+
+            // Read the desired visibility from the switch.
+            bool visible = ReferenceLayersSwitch.IsToggled;
+
+            // Set each reference layer to that visibility.
+            foreach (var layer in basemap.ReferenceLayers)
+            {
+                layer.IsVisible = visible;
+            }
+        }
```
imalcolm1: There's an issue here (and on WPF and WinUI) where if ToggleReference layers is off and you switch between basemaps or between manual and automatic mode the layers will sometimes still show, particularly on the default basemaps. The fix seems to be to load the basemap before updating the layer visibility, which could either go here or somewhere in UpdateBasemapAsync
```suggestion
        private async Task ApplyReferenceLayerVisibility(Basemap basemap)
        {
            // Skip if the basemap has no reference layers.
            if (basemap?.ReferenceLayers == null) return;

            // Read the desired visibility from the switch.
            bool visible = ReferenceLayersSwitch.IsToggled;
            
            // Ensure the basemap is loaded
            await basemap.LoadAsync();

            // Set each reference layer to that visibility.
            foreach (var layer in basemap.ReferenceLayers)
            {
                layer.IsVisible = visible;
            }
        }
```

src/MAUI/Maui.Samples/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
imalcolm1, line 1: After clicking either of the open settings buttons on android, the app subsequently takes me to the same page no matter which one I click afterwards. This could be an android/samsung bug. Also, for the contrast settings link it navigates to general accessibility settings rather than the specific contrast settings page, but that one does seem like an android limitation.

src/WinUI/ArcGIS.WinUI.Viewer/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
```diff
@@ -0,0 +1,300 @@
...
+            }
+
+            // In dark mode UISettings returns black for Background; in light mode it returns white.
+            Color background = uiSettings.GetColorValue(UIColorType.Background);
+            return background.R > 127;
+        }
+
+        // Windows theme or high-contrast changed. ThemeSettings.Changed fires on the UI thread
+        // and covers both theme and high-contrast transitions.
+        private void OnThemeSettingsChanged(ThemeSettings sender, object args)
```
imalcolm1, line 280: This doesn't seem to trigger when I switch light/dark theme. It works fine for high contrast though

imalcolm1 (changes requested):
A couple of things but otherwise looks really good!

src/WPF/WPF.Viewer/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
praveenaak: Root cause turned out to be a stale cache interaction between the viewer's `Sample_Unloaded` handler and this sample's `_lastAppliedChoice` cache.` Sample_Unloaded` nulls` MyMapView.Map` (grey grid), and on the reattach `_lastAppliedChoice` still matches the HC choice so `UpdateMapForAppearanceAsync` early returns and never rebuilds. Toggling manual and back works because that path resets the cache. Fix is to reset `_lastAppliedChoice = null` in OnUnloaded too. The earlier workaround was SwapMapAsync - a null-flush-reassign helper that tried to force a clean repaint but wasn't actually addressing the real cause (stale cache).

src/MAUI/Maui.Samples/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
praveenaak: This does look like an OS-side bug and a limitation. Workaround for now: close the settings page from the background before clicking the other button.

src/WinUI/ArcGIS.WinUI.Viewer/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
praveenaak: added the fix

src/MAUI/Maui.Samples/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
```diff
@@ -153,13 +153,16 @@ private static async Task<PortalItem> LoadPortalItemAsync(string itemId)
...
+        private async Task ApplyReferenceLayerVisibility(Basemap basemap)
         {
             // Skip if the basemap has no reference layers.
             if (basemap?.ReferenceLayers == null) return;
 
             // Read the desired visibility from the switch.
             bool visible = ReferenceLayersSwitch.IsToggled;
+            
+            // Ensure the basemap is loaded
+            await basemap.LoadAsync();
```
imalcolm1, line 165: I think WinUI needs the same fix as here for the reference layers still showing problem. WPF seeeems to be working without the fix, but I'm not sure we should trust that to hold, so it could probably get the same treatment.

imalcolm1 (approved):
The reference layers thing is still a bit off, at least on WinUI, but otherwise looks good to go

src/MAUI/Maui.Samples/Samples/Accessibility/UpdateBasemapForContrastAccessibility/UpdateBasemapForContrastAccessibility.xaml.cs
praveenaak: Missed this in the previous iteration. Added the same fix to WinUI and WPF.

PR #1824, opened 2026-08-04
Handle search event handler and dismiss keyboard

OAzmitia:
# Description

When searching for samples based on keywords, when the list has more than a few items the keyboard won't allow you to view the bottom samples.

In the video below you see that after searching for "marker" it is impossible to select the bottom two results on iOS. Selecting the "search" button does not allow you to dismiss the keyboard.

|**Issue**|**Fix**|
|-----|-----|
| <video src="https://github.com/user-attachments/assets/6fcdbe7e-fae0-47fe-b8cf-09a77cb4ca50"/> | <video src="https://github.com/user-attachments/assets/c4ea7a13-8290-41c8-9316-e298ab484907" /> |

## Type of change

- Bug fix

## Platforms tested on

- [x] MAUI WinUI
- [x] MAUI Android
- [x] MAUI iOS
- [x] MAUI MacCatalyst

## Checklist

- [x] Self-review of changes


src/MAUI/Maui.Samples/Views/SearchPage.xaml
```diff
@@ -13,10 +13,13 @@
                    Grid.Column="0"
                    ios:SearchBar.SearchBarStyle="Minimal"
                    Placeholder="Enter search term"
-                   SearchCommand="{Binding PerformSearchCommand}"
```
imalcolm1, line 16: minor: It doesn't look like `PerformSearchCommand` is used anywhere else, so we can probably also remove the `[RelayCommand]` attribute on `PerformSearch()` in the view model (and make it `private` while we're at it, not sure why it didn't have a scope keyword before).

imalcolm1 (approved):
Nice fix! We'll wait for the release to merge

src/MAUI/Maui.Samples/Views/SearchPage.xaml
OAzmitia: Actually, we do use it on line 20 and 22. We use it when the user executes either search button behavior or they stopped typing.

imalcolm1: Oh my bad, I must have searched in v.next rather than this branch. Good to go then!

PR #1826, opened 2026-08-04
Fix build checks

imalcolm1:
# Description

The commits are everywhere for this one, but the main changes are as follows:
1.  Made the generator project keep its framework and rid properties local only so they can't be overridden when the main projects import it
2. Updated the Maui and WinUI builds to use `dotnet build` since only WPF needs the full developer command line dependencies and it makes the commands simpler
3. Made the WPF build only target x64 since the store build project was trying to build x86. We could also build arm64 but that doesn't seem worthwhile
4. Updated the WinUI build command to set the output directories to the repo root to avoid errors from excessively long file path errors

## Type of change

- Bug fix

## Platforms tested on

- [x] WPF
- [x] WinUI
- [x] Maui Windows
- [ ] Maui Android
- [ ] Maui iOS
- [ ] Maui Catalyst

## Checklist

- [x] Self-review of changes
- [ ] Screenshots are correct size and display in description tab (800 x 600 on Windows, MAUI mobile platforms use the MAUI Windows screenshot)


src/WPF/WPF.StorePackage/ArcGIS.WPF.StorePackage.wapproj
```diff
@@ -55,7 +55,7 @@
     <None Include="ArcGIS.WPF.StorePackage_TemporaryKey.pfx" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\WPF.Viewer\ArcGIS.WPF.Viewer.Net.csproj" AdditionalProperties="TargetFramework=net8.0-windows10.0.19041.0">
```
imalcolm1, line 58: This change is not absolutely necessary, but it means we won't have to update this to net10.0 later

praveenaak (approved):
lgtm!

